PR #84, opened 2014-09-05
Update bld/wv/c/wndsys.c

beketata:
It was a mistake to use WndGadgetArraySize variable (which is actually equal to 39) in those two "for" cycles dealing with WndGadgetHint array (which is actually contains 19 elements).

### DIFF
--- a/bld/wv/c/wndsys.c
+++ b/bld/wv/c/wndsys.c
@@ -118,6 +118,8 @@ gui_resource WndGadgetArray[] = {
     #undef pick
     { BITMAP_SPLASH, "splash" },
 };
+
+int WndGadgetHintSize = ArraySize( WndGadgetHint );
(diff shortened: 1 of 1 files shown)


iainnicol:
Hi @beketata. We are trying to relicense Open Watcom under a more mainstream licence. Are you okay to relicense your contributions under [Apache 2.0 + LLVM Exceptions](https://llvm.org/LICENSE.txt)? It's the same licence used by LLVM and Clang.

If you want to why we want to do this, I wrote up the background [on the wiki](https://github.com/open-watcom/open-watcom-v2/wiki/Relicensing-effort).

I wasn't able to contact you by email, but with any luck you'll see this GitHub message :-).

sskras:
Repeating the ping!

PR #132, opened 2014-12-30
Fixes for the unix version of _bgetcmd

ArmstrongJ:
The `_bgetcmd` function was failing under extremely common circumstances under Linux.  Specifically, it would never capture the actual executable name (it was starting with the first argument).  Additionally, it now properly quotes arguments with spaces when reconstructing the command line.

### DIFF
--- a/bld/clib/environ/c/gtcmdunx.c
+++ b/bld/clib/environ/c/gtcmdunx.c
@@ -48,12 +48,12 @@ _WCRTLINK int (_bgetcmd)( char *buffer, int len )
     int     i;
     char    *word;
     char    *p     = NULL;
-    char    **argv = &_argv[0];
+    char    **argv = &_argv[1];
     char    *space;
 
     --len; // reserve space for NULL byte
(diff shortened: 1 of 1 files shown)


ArmstrongJ:
Sorry about that, @jmalak.  The latest commit actually does what `_bgetcmd` is supposed to do.  The original code had two major issues.

1. If there were no arguments and you passed a buffer, the buffer would return with garbage.  If the buffer is sized properly, it should actually return a string of zero length that is properly null-terminated.  That was the cause of issues I was seeing when trying to use code that called the command.

2. There were multiple places in the code where the remaining buffer length was not verified before adding more characters.  I've added a significant number of additional checks (making the code a bit ugly).

The earlier commit's quoting of problematic arguments is still present, as it should be.

jmalak:
I reworked your fix as da3087bc76fa88d6414569402d4b93813c069c8d to handle properly all problems.
Thanks for your contribution.

PR #133, opened 2015-01-05
Implementations of getline and getdelim.

ArmstrongJ:
The following commit adds getline and getdelim, both POSIX functions, to all Open Watcom platforms.  Documentation is included.

I'm not sure how people feel about these POSIX functions being included on other platforms.  I don't see a real problem with it, however.


jmalak:
Thanks for your contribution.

PR #135, opened 2015-01-11
Implementations of getpwent/getpwuid/getpwnam

ArmstrongJ:
This commit includes implementations of standard POSIX functions for accessing the password database (on Linux).  Documentation is also included.  The headers for these functions were already present.


jmalak:
Thanks for your contribution.

PR #140, opened 2015-02-18
Support for Fortran in owcc wrapper.  Improvements to command-line handling in owcc and Fortran compilers.

ArmstrongJ:
This commit (squash of 5) adds a few new Fortran-centric features, notably:

* Drastically better command-line handling in the Fortran compiler, the "wfl" driver, and the owcc driver by using spawnvp rather than erroneously using spawnlp
* Support for the Fortran compiler in the POSIX wrapper driver (owcc) when a .f or .for file is encountered
* Addition of compiler options to the Fortran compiler for compatibility with other Open Watcom tools (-zq, -fr, -i are now valid)
* Support for combinatorial optimization options in the Fortran compiler (like "-ohxl+" is now processed properly, just like our C compilers)
* Support for spaces in the objects and linked executable name when using wfl or owcc

With support for the Fortran compiler in the POSIX wrapper, we can start to look at providing proper support for complex math packages that expect a POSIX-like command line.  

The command line argument changes were necessary because the Fortran compiler was mis-parsing options due to its reconstruction of the command line prior to parsing.  Using spawnlp was causing the Fortran compilers to treat, for example, "-q -er" as a single option on Linux (by re-adding quotes when constructing the command line), causing all sorts of issues in the compiler.  Passing arguments one-by-oe from wfl and owcc properly constructs the argument list now when launching the child processes.

This pull request is rather large, but adding one thing required fixing another, which required fixing another...

### DIFF
--- a/bld/f77/wfc/c/option.c
+++ b/bld/f77/wfc/c/option.c
@@ -69,6 +69,7 @@ extern  uint            DataThreshold;
 /* Forward declarations */
 static  int     GetDirective( char *buff );
 static  void    ScanOpts( char *buff );
+static  void    CompoundOptOption( char *buff );
 
 // Compiler directives
 
@@ -509,7 +510,6 @@ static  opt_entry       *GetOptn( char *buff, bool *negated ) {
     return( NULL );
 }
 
-
 void    CmdOption( char *buff ) {
 //===============================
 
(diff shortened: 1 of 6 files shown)


jmalak:
It is fine. wfl was very old piece of code with minimal support for Linux.
I did a small enhancement in the past, but your work is much better. 
Thanks for adding Fortran stuff to owcc.

PR #147, opened 2015-03-12
Implementation of POSIX timers

ArmstrongJ:
Although present in our headers, we didn't have an implementation of the POSIX timers.  This pull request implements (with docs) the following:

* timer_create
* timer_delete
* timer_settime
* timer_gettime
* timer_getoverrun

All functions appear to be working on a Linux kernel that supports timers.

### DIFF
--- a/bld/clib/h/sys386.h
+++ b/bld/clib/h/sys386.h
@@ -298,6 +298,12 @@
 #define SYS_sched_setaffinity   241
 #define SYS_sched_getaffinity   242
 #define SYS_set_thread_area     243
+/* ... */
+#define SYS_timer_create        259
+#define SYS_timer_settime       260
+#define SYS_timer_gettime       261
+#define SYS_timer_getoverrun    262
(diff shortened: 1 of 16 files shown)


jmalak:
Thanks for your contribution

jmalak:
Hi Jeff, 
I setup your profile to have write access to Open Watcom V2 repositories that you can submit your change directly without change request.
Jiri

ArmstrongJ:
Jiri,
I'm honored, thank you! I promise only to use my commit powers for good.

jmalak:
Now you understand OW well and there is no reason not to give you write access. I'm not a big proponent of restrictions in programming because it's creative work and it needs freedom. When we make a mistake we can fix it, it is not end of the world. Of course I am not saying do mistakes. I am doing more and more mistakes because I am older and I am not a professional programmer. It is essential to correct errors. Sometimes it is not as fast as everyone thinks because work on OW we do in our free time.

Jiri

PR #153, opened 2015-03-24
Fixed references to PAGE_SIZE macro in help compilers/viewers

ArmstrongJ:
This commit should eliminate using the possibly problematic term `PAGE_SIZE` within some of the help compilers and viewers.  It appears that these changes are building fine.


jmalak:
It is OK for me.

jmalak:
I already fix it.

PR #154, opened 2015-04-03
Implementation of nl_langinfo

ArmstrongJ:
I've included commits that create a basic implementation of `nl_langinfo` for Open Watcom.  Rather than a direct commit, I decided on a pull request in case anyone has any suggestions if they don't like how things are managed.  Basically, the function provides a simple U.S.-centric implementation of nl_langinfo (English spellings and US-standard numerical separators) that's better than nothing.  It also suggests that our supported codeset is UTF-8, which is safest for any ASCII string functions while not overly constricted.  

I also included a minor change to `setlocale` that allows the "POSIX" locale to be specified.  Per the POSIX standard, it is the same as the "C" locale, so it is implemented merely as a check-and-substitution.


jmalak:
Hi, 
I would like to discuse macro
#define BASE_CODESET "UTF-8"
with regards to current OW clib implementation.
I am not expert on POSIX international support, but generaly "UTF-8" is multi-byte encoding not single byte.
I think that OW multi-byte support for Linux is minimal.
Therefore I think the macro should be defined as
#define BASE_CODESET "ISO-8859-1".
I know that for POSIX "portable" codeset is no difference between "UTF-8" and "ISO-8859-1", but processing can be different as soon as characters are outside "portable" codeset.
It is only my opinion and need not be correct.

ArmstrongJ:
The codeset value was one reason I wanted to do a pull request with these commits.  I decided on UTF-8 because, while it is multibyte, it is technically compatible with our standard library string functions.  UTF-8 strings are still terminated by a single null character, and functions like strlen and strcpy will still work fine.  There could be issues if someone were to, say, attempt to use strchr with a multibyte character, but strstr, for example, would continue to work fine even with UTF-8 search terms.

I don't actually have a problem, though, with switching to ISO-8859-1.  Anyone else have some input?

jmalak:
I agree that target value should be "UTF-8", but it will require some work on clib for Linux to properly handle "UTF-8". QNX stuff should have this support, but I don't know how complete it is.

pchapin:
I also agree that long term, UTF-8 is a better place to be than ISO-8859-1. However, I don't know all the implications of going multibyte so I can't comment on that. ISO-8859-1 is probably fine for now.

jmalak:
Sorry, correct value is "US-ASCII", not "ISO-8859-1".

rdos314:
RDOS target assumes all strings are UTF-8, and the wide-string support is disabled.

jmalak:
For me it looks OK.

ArmstrongJ:
Thanks everyone for the input!

PR #177, opened 2015-06-10
Fix crash in wlinkd.dll when running wlink on OS/2

groessler:
There were some strange '#ifdef __OS2__' parts in wl/c/ideentry.c, which caused wlink to crash.
I've disabled them for now (the IDE crashes anyway on startup).
I've successfully compiled 16 and 32 bit OS/2 test programs and 16 bit DOS program on a OS/2 host with this change.

### DIFF
--- a/bld/wl/c/ideentry.c
+++ b/bld/wl/c/ideentry.c
@@ -56,7 +56,7 @@ static IDECBHdl         IdeHdl;
 static IDEInitInfo      InitInfo;
 static IDECallBacks     *IdeCB;
 
-#if defined( __OS2__ )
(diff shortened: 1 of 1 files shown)


jmalak:
Thanks for your contribution.

groessler:
Thanks for inclusion. I omitted an important part in my description: The crash happened due to a duplicate call to FiniMsg().

jmalak:
It is OK. It was historical code which is included now in InitSubSystems and FiniSubSystems.
Thanks for checking on OS/2. We don't have anybody who is using/testing OW v2 on this OS regulary and contribute by bug report or by fix.

PR #178, opened 2015-06-18
add extern "C" to make the header work in C++

Azarien:
Without extern "C" linker couldn't resolve functions in cwdll.lib


jmalak:
Thanks for your contribution.

PR #188, opened 2015-07-24
Win32s stack limits fix 20150724

joncampbell123:
I think I have a more universal fix for the stack issue in Windows 3.1.

This patch keeps the low point adjustment logic but prevents the low point from going too far so that there is valid stack to work from. I explain in this patch that the main openwatcom-v2 branch cannot produce working Win32s binaries with stack checking enabled because the default stack size set by wlink for win32s is 64KB (look at the EXE's PE header to see what I mean), yet __init_stack_limits() sets the low point to the base of the stack region + 72KB which puts the low point past the top of the stack and the program crashes immediately with a "stack overflow" error. This fix keeps the low point calculation but prevents it from going too far so that the program has a chance to run. With this patch applied, my branch of openwatcom-v2 is able to produce working Win32s binaries that run under Windows 3.1.

I want to clarify the stack issue only happens under Windows 3.1 because when the PE header lists a 64KB reserved stack size, Windows 3.1 gives the Win32s program exactly that, and that's where we get the problem with the low stack point in __init_stack_limits() set to base + 72KB.

EDIT: Go ahead and ignore the parts of this patch that change the *.mif file to disable building documentation. What matters are the changes to bld/clib/startup/c/stklmwnt.c


joncampbell123:
Hold on... I see a problem. I made the classic mistake of using (a + b) / 2 without regard to integer overflow. I need to change it to a + ((b - a) / 2) to avoid problems if the stack base or top end up just past the 2GB mark. I will fix it on my end and resubmit the pull request.

jmalak:
Please, could you review this pull request and create new one or remove it.

PR #203, opened 2015-10-11
Linux issues solving

seyko2:
Some corner problems: 
  * if terminfo base can not be found
  * if working on linux console with utf8 (not xterm)
  * if trying to build WATCOM with gcc 3.4.6 (or with other not c++0x compatible compiler)
  * allow to start build WATCOM by running build.sh w/o doing anything else (tuning)


jmalak:
Many Thanks for your contribution.
Submited in git repository as 63420c743137d8bf49952cb581fa3b2b43de0367

PR #207, opened 2015-11-01
vi on unix: show (turn on) an edit window border

seyko2:
vi from the watcom 1.9 has this. I don't found
    why it disappears in the current version.


jmalak:
It is now fixed. There were two mistakes which affect it.
- mistake in ASSIGN command processing
- mistake in SET commad processing of GADGETSTRING item

PR #241, opened 2016-03-05
Implementation of POSIX semaphores for Linux systems

ArmstrongJ:
This pull request implements a functioning semaphore system on Linux.  These semaphores have also been integrated into the _mthread_ module so that Linux threads should now be more-or-less completely implemented there.

One problematic point is that the POSIX header file for semaphores should be named _semaphore.h_, which obviously doesn't conform to Open Watcom's 8.3 file name limitations. In the makefile, I did implement a custom rule just to create this file from _sema4.mh_.

The commit is big enough (plus the custom makefile rule...) that I thought a pull request might be the best course of action.


jmalak:
We should hold 8.3 file name format for all files in OW tree.
For header files we have defined _ialias.h file where similar files are defined as alias for header file real name.
We should define header file as semaphor.h and add folowing line to _ialias.mh.
#pragma include_alias( <semaphore.h>,        <semaphor.h> )

sema4.mh should be renamed to semaphor.mh, that generated header file will be semaphor.h

ArmstrongJ:
@jmalak thanks for the tip!  I wasn't sure how that was handled.

Also, these commits will only work for x86 Linux right now.  The file __atomic.c in the original commit contains a small assembly call that would need implementation on other platforms.

jmalak:
For x86 code, it uses cmpxchg command and if I remeber properly then this is not available on all 32-bit Intel processors. We should handle it somehow.

ArmstrongJ:
The `CMPXCHG` instruction was introduced on the 486. That leaves only 386 chips where I won't work. I'm open to suggestions on how to address it.

jmalak:
In the past I introduced bld/clib/startup/c/rtcntrl.c file which we could use for adding new flag setup during linux CRTL startup. We could set new bit in control variable which identify presence of cmpxchg instruction by example. If cmpxchg will be not available then appropriate CRTL fatal error message should be displayed and application exit.

jmalak:
You should also update deps.mif in hdr project for semaphor.mh dependencies.

ArmstrongJ:
@jmalak Thanks for merging!  I was still looking for a possibly more elegant solution to the "detect if CMPXCHG is present" issue, but merging now is fine.  The whole issue might not matter much since:

- The implementation requires Linux futexes, introduced in kernel 2.6
- All 386 support was dropped starting with kernel 3.8

I don't think our goal is to support legacy Linux, but we can still look for a nice way to detect a >=486 chip.  I was actually thinking raising an `ENOSYS` on `sem_init` if the CPU wasn't supported (<486 or other architecture) might be the most elegant solution.

jmalak:
I did simple checking for 486 CPU by ID bit, even if some 486 CPU doesn't support it.
But it was 486 SX and some earlier 486 DX. By CPUID we can get more CPU details if we need it.
I think now we have good starting point and code can be improved

rdos314:
I think distinguishing 386 from 486 is done by trying to change bit 18 in EFLAGS. Then trying to change bit 21 in EFLAGS will indicate if CPUID is supported. Something like below:

    mov ds:cpu_type,3
    mov ds:cpu_feature_flags,0
    mov ds:cpu_ext_feature_flags,0
    mov ds:max_cpuid,0
;
    pushfd
    pop eax
    mov ecx,eax
    xor eax,40000h
    push eax
    popfd
    pushfd
    pop eax
    xor eax,ecx
    jz cpu_done
;
    mov ds:cpu_feature_flags,1
    mov ds:cpu_type,4
    mov eax,ecx
    xor eax,200000h
    push eax
    popfd
    pushfd
    pop eax
    xor eax,ecx
    je cpu_done
;
    mov eax,0
    cpuid
    mov ds:cpu_vendor+12,0
    mov dword ptr ds:cpu_vendor,ebx   
    mov dword ptr ds:cpu_vendor+4,edx   
    mov dword ptr ds:cpu_vendor+8,ecx  

cpu_done:

PR #244, opened 2016-05-08
convert wgml.bat to CR-LF line-ending

stsp:
Some command.com's do not understand the unix's LF lines.


jmalak:
Thanks for your contribution

PR #245, opened 2016-05-08
fix crash on build [fixes #184]

### LINKED ISSUE #184: Build errors on FreeBSD 11
Evening,

I experience some build errors with OpenWatcom on FreeBSD 11-Current.

The first of note is that it cannot reach builder due to being unable to find gcc. When switched to clang mode it instead complains about being unable to find libsupc++ and libstdc++; replacing references to those with -lc++ produces problems as well.

The next is that a file which makes reference to off_t (it #defines something as off_t) appears to fail to include unistd.h, which defines off_t on FreeBSD. This fails at some point in the build process where it is included by a C file that does not include unistd.h before it.

Finally, I am stalled at this:
```**** BUILD rule
============= 19:56:10 /ws/Misc/open-watcom-v2-master/bld/browser =============
=========== 19:56:10 /ws/Misc/open-watcom-v2-master/bld/browser/gen ===========
======= 19:56:10 /ws/Misc/open-watcom-v2-master/bld/browser/brg/dos386 ========
======== 19:56:10 /ws/Misc/open-watcom-v2-master/bld/browser/brg/nt386 ========
======= 19:56:10 /ws/Misc/open-watcom-v2-master/bld/browser/brg/os2386 ========
========== 19:56:10 /ws/Misc/open-watcom-v2-master/bld/browser/nt386 ==========
dosbox -noautoexec -c "mount c /ws/Misc/open-watcom-v2-master/docs" -c "mount d ." -c "mount e /ws/Misc/open-watcom-v2-master/bld" -c "d:wgml.bat" -noconsole
DOSBox version 0.74
Copyright 2002-2010 DOSBox Team, published under GNU GPL.
---
CONFIG:Loading primary settings from config file /home/david/.dosbox/dosbox-0.74.conf
MIDI:Opened device:none
DOSBox switched to max cycles, because of the setting: cycles=auto. If the game runs too fast try a fixed cycles amount in DOSBox's options.
Abort
mv: rename TMP.PTF to wbrw.ptf: No such file or directory
Error(E42): Last command making (compile_gml_to_xxx;wbrw.h) returned a bad status
Error(E42): Last command making (wbrw) returned a bad status
Error(E42): Last command making (/ws/Misc/open-watcom-v2-master/docs/nt/wbrw.h) returned a bad status
<pmake -d build         -h> => non-zero return: 512
Build failed```

This is documentation related, it seems; that I `export OWDOCBUILD=0` in my local setvars.sh copy does not appear to affect this issue, and neither does setting OWDOCBUILD=1 help.

stsp:


jmalak:
Thanks for your contribution.

Fix for  bld/plusplus/c/scope.c aplied.

Fix for bld/plusplus/c/convctl.c need review.

stsp:
> Fix for bld/plusplus/c/convctl.c need review.

Yes.
Let me clarify what actually happens.
When https://github.com/stsp/open-watcom-v2/blob/fix_crash/bld/plusplus/c/convctl.c#L1259
retn = classPtrConversion( a_expr, tgt_type, conversion );
returns something other than CNV_OK_CV, the "retn"
value is not fixed-up to CNV_OK. In this case the branch
https://github.com/stsp/open-watcom-v2/blob/fix_crash/bld/plusplus/c/convctl.c#L1267
does not enter, and "expr_type" is left uninitialized.
However, it is still used on the branch
https://github.com/stsp/open-watcom-v2/blob/fix_crash/bld/plusplus/c/convctl.c#L1291
namely, here:
https://github.com/stsp/open-watcom-v2/blob/fix_crash/bld/plusplus/c/convctl.c#L1300
Then the crash.

Please note that there is an alternative code disabled:
https://github.com/stsp/open-watcom-v2/blob/fix_crash/bld/plusplus/c/convctl.c#L1237
It looks safer as it initializes "expr_type" in any case,
but I guess this was considered slow and some "optimization"
was applied instead, that now leads to crash.

Please make whatever fix you think necessary.
I only tracked and patched the one particular return
value that I observed, but you should know a better fix.

jmalak:
Thanks for your analysis.
I must look into code deeper because I am not sure if classPtrConversion should only report  if conversion is OK or other exception or if it should realy convert class pointer because rule for class pointer can be different from member pointers processing.
No problem on 32-bit only if segmented architecture is used then problem appears.
Anyway there is some strange handling of errors in debug version of C++ compiler that report problem with virtual machine stack even if non-debug version pass it.

stsp:
Ok but this bug is quite "fatal" as it prevents
the build process from succeeding.
So maybe you can at least temporary apply
some (my?) work-around, so that people can
at least build the thing...

jmalak:
I applied what was transparent. But problem with Class pointers conversion is little bit deeper.
You created fix for crash only, but it doesn't fix solve source of problem.
By example I have no problem on OpenSuse. 
It looks like some type of memory problem which is masked by used compiler or used run-time library.
I am still investigating appropriate parts of C++ compiler.

jmalak:
Anyway you can overcome this problem by builder command
builder -i build
This continue build even if error is reported.

jmalak:
Or next posibility is
builder build cpu_i86 .not
which exclude i86 target from build

stsp:
> which exclude i86 target from build

No, I need the DOS target.

> By example I have no problem on OpenSuse.

But for that I already created a script for you that
will help you to reproduce. Here it is again:
https://github.com/open-watcom/open-watcom-v2/files/253679/a.sh.txt

> It looks like some type of memory problem which is masked by used compiler
> or used run-time library.

But as I have explained, the memory problem is
that "expr_type" is left uninitialized. And I described the
exact code path how that happens. There is also the
disabled code that doesn't have this problem - please
take a look into that one too.
I am not saying my fix is best, but the most important
thing right now is to have "expr_type" initialized.

jmalak:
Please, could you confirm if problem is fixed by latest changes.
I understand that source of this problem is still not fixed because I can not reproduce it on my Linux installations (Open Suse and FreeBSD), but I fixed some discrepancies (potential overflow) in virtual stack which can have impact to this problem. Anyway 64-bit build (Unix LP mode) of C++ compiler is buggy, don't pass many tests.

Please what platform you use to build OW?

stsp:
> Please, could you confirm if problem is fixed by latest changes.

Yes, looks better. In particular, the crash stops happening
after commit 4f92b797, which looks like a cleanup? Strange.
Thanks!

> I understand that source of this problem is still not fixed because I can not reproduce
> it on my Linux 

Any reason you are not using the script I provided?

> Please what platform you use to build OW?

Fedora-24.

jmalak:
The problem is that C++ compiler code is oriented to speed and some path have uninitialized variables under some conditions. Therefore if we change something then similar problem can appear.
I tested your script but it is generate problem where it is not. I preferer logical (static) analysis because I also lerning how is C++ compiler designed. There are many problems related to 64-bit host and will require more effort. I will try to install Fedora to VM to be able check in the future.

stsp:
> I tested your script but it is generate problem where it is not.

valgrind is known to be extremely, exceptionally precise.
With my script, it shows 2 places for every problem: where
the uninitialized var was created, and where it was used.
By "used" he means really used. For example, if the uninitialized
var is just copied to another, he keeps silence. But if it was used
in the evaluation of the conditional under "if", then it warns.
Or if it was an uninitialized pointer and was used to read/write
memory, then it warns.
If you want, I can take 1-2 problems he reports, analyze them
and produce the patch - that's how much I trust valgrind. :)

Do you think 4f92b79 could really fix anything at all, or was
my testing somehow bad?

jmalak:
I don't mean that valgrind is poor tool, but for our problem with C++ compiler it is only auxiliary information.
By example crash was caused by not initialized "expr_type" (it was properly detected by vagrind).
Source of problem is that the condition why "expr_type" was not initialized was wrong. If compiler work properly then this can not happen for class pointers etc. In C++ compiler is more similar bugs, but it requires rather logical analysis (and fixing of these logical mistakes) that simple fixing of uninitialized variables because it doesn't guarantie compiler proper functionality, only fix crash.  Fix 4f92b79 is one of many required, it is start not end. But this fixing is time consuming therefore any help is welcome.

stsp:
OK, lets get this closed then.
Btw, is it possible to use the parallel build on multicore?
./build.sh -j 16
doesn't seem to give anything good.

PR #246, opened 2016-05-08
fix prototypes

stsp:


jmalak:
Thanks for your contribution

PR #253, opened 2016-05-20
build.sh: use bash

stsp:
Many distributions (ubuntu, debian) use dash instead of bash.
dash gives the following error:
./build.sh: 9: ./build.sh: source: not found
./build.sh: 20: cd: can't cd to /wmake

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -6,7 +6,7 @@
 # Expects POSIX or OW tools.
 
 if [ -z "$OWROOT" ]; then
-    source ./setvars.sh
+    . ./setvars.sh
 fi
 
 OWBUILDER_BOOTX_OUTPUT=$OWROOT/bootx.log


groessler:
bash is not always installed in /bin. On FreeBSD for example it's in /usr/local/bin.

I suggest to use

#!/usr/bin/env bash

groessler:
A better fix would be to get rid of the bashisms.

use

. ./setvars.sh

in lin e #9.

ArmstrongJ:
We shouldn't assume the presence of bash.  I agree with @groessler.  We should fix the script error, not force bash.

stsp:
OK, done.

XVilka:
You can try to run [shellcheck](https://github.com/koalaman/shellcheck) on all shell scripts in openwatcom to find incompatibilites, errors, etc.

stsp:
The script is too small, I am just checking the full
build now.
Does anyone know how to start a parallel build on multicore?

ArmstrongJ:
I'm pretty sure Open Watcom build system doesn't support parallel building.  Did something make you think it was possible?

stsp:
Something make me _wish_ it was possible. :)
And that something is: hours of build time on the PC that
could do it in 5 minutes if a parallel build was used...

jmalak:
Thanks for contribution, appliied.

PR #255, opened 2016-05-25
clean up dosemu setup

stsp:
dosemu issues the useless warnings and waits for ENTER to
be pressed. Instead of messing around its configs, we can
fortunately just use the "-quiet" option to suppress all
stupid things.


stsp:
Note: I don't know if this works.
I have submitted it so that the automated check
can check whether this is right or wrong.
If check succeeds, then fine.

jmalak:
thanks for contribution

PR #256, opened 2016-05-29
Set OWDOCBUILD=1 by default, add IFDEFs to builder.ctl

joncampbell123:
The IFDEFs allow you to set OWDOCBUILD=0 to skip building the documentation, browser, and IDE and anything else that depends on WGML to build.

### DIFF
--- a/bld/builder.ctl
+++ b/bld/builder.ctl
@@ -84,8 +84,10 @@ cdsay .
 [ INCLUDE <OWSRCDIR>/bmp2eps/builder.ctl ]
 # Build ide2make tool for idedemo
 [ INCLUDE <OWSRCDIR>/wclass/builder.ctl ]
+[ IFDEF <OWDOCBUILD> 1 ]
 [ INCLUDE <OWSRCDIR>/ide/lib/builder.ctl ]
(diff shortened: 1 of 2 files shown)


jmalak:
I am sorry, this change is not acceptable.
Purpose of OWDOCBUILD variable is to build/no-build documentation only it has no impact on build OW tools.
It can not be used to handle sub-set of OW tools for building any way.

jmalak:
You must add some new variable and document its purpose, but it can not be OWDOCBUILD variable.

ArmstrongJ:
I agree with @jmalak.  People do regularly set `OWDOCBUILD=0` so they can build everything but documentation since it tends to be time-consuming. 

I'd suggest adding a variable `OWGUIBUILD` that would enable/disable building any of the GUI tools, and set it to 1 by default.

Also, try framing this as "disabling building all the GUI tools" rather than just "avoiding WGML." This change is actually a good idea as long as it is explicit (a new variable is used in `setvars.sh`) and doesn't change the default behavior. With that minor change to using a new variable, I think this change should and will get pulled into master.

joncampbell123:
@ArmstrongJ Will do.

joncampbell123:
Try #2 submitted.

PR #257, opened 2016-05-31
OWDOCBUILD to enable documentation, and OWGUIBUILD to enable the IDE.

joncampbell123:
also add message to build.sh to remind what to do after build.


jmalak:
Reimplemented proper way (using pmake guitool tag) to be able handle realy only GUI version of tools. Some projects have GUI and Character mode versions. db1f8af9c0fe4d5d0db2c7d0f462f2d6b9100704 , 7af1d6c83e7a49a96bd473976851b359ff915caa ,  7127bad141f4f05552e51bb87af0791c33b3808d

PR #281, opened 2016-09-29
New, original implementation of POSIX threads for Linux platform

ArmstrongJ:
This rather large pull request implements the entire POSIX threads API for the Linux platform.  The pull request is a single, squashed commit based on [the posix-thread-1 branch of ArmstrongJ/open-watcom-v2](https://github.com/ArmstrongJ/open-watcom-v2/tree/posix-thread-1), a series of 29 commits that have been rebased to current master.

The threading implementation relies on two major components: Open Watcom's internal threading system (bld/clib/mthread) and POSIX semaphores.  Mutexes, for example, rely on a pair of semaphores for handling locking in a reliable manner.  POSIX threads are accounted for internally in a linked list within the process separate from Open Watcom's internal accounting of threads.  Most other blocking functionality derives from mutexes (with the exception of spinlocks, which rely on some Linux-specific atomic operations).

Although this implementation might be simplistic, it should be easy to port to any other supported platform as long as POSIX semaphores are ported first.  Thread launching is handled by Open Watcom's mthread library, and the code makes no assuumptions about the platform.  If so desired, someone could implement POSIX semaphore wrappers around Windows or OS/2 semaphores, and this threading implementation should "just work" (although some attributes related to scheduling won't).

This pull request is rather large, but it is almost entirely new code.  It shouldn't cause any side effects.

I'm open to comments!  As I stated, this implementation is rather simplistic, but it should work properly.  I have tested it to some degree, obviously, and it seems to be working.

PR #283, opened 2016-10-07
Updated implementation of POSIX threads for Linux platform

ArmstrongJ:
This pull request replaces #281 by incorporating current changes on master.  Some commits originally incorporated into #281 have been eliminated due to additional changes on master that negate some merged code from the original branch.


ArmstrongJ:
There are still some minor issues in mthread.c (still not sure how I was successfully building before, but...).  I'll look at it more tomorrow.

ArmstrongJ:
@jmalak I appreciate all these quick updates!  I originally added the file list locking to try to lock console input/output, but I think it was unnecessary (did we change a syscall in master at some point related to this?).  I don't remember the specifics any longer, but console I/O is working just fine.

jmalak:
No problem. Your stuff is great that now we will be able to compile single/multi-threaded applications on Linux and we can cleanup code which was half no-thread/single-thread/multi-thread code combination.
Unfortunately it is great that we have POSIX thread API in CRTL.
As soon as you thinking it is ready for integration to OW2 master branch then integrate it.

ArmstrongJ:
I've proceeded with the merge after running a few simple checks. There is still a bit more work I'd like to do with respect to `pthread_join` and `pthread_detach`, but it should be working at this time.

Thank you for all the help and additional runtime library changes!

jmalak:
I did investigation of core multi-threaded code for Linux and there is one fatal problem now.
There is missing stuff for handling TLS on Linux (structure thread_data used by OW for this purpose). The code is still single-threaded only.
Therefore any thread local variables crash if more than one thread will be starting by _beginthread.
If I understand properly then Linux uses GS register to access TLS, but it must be initialized somehow.
Do you have any idea how to handle it under Linux?

jmalak:
This problem is mainly related to Fortran and C++ run-time libraries which handle TLS, extend it for structures used by appropriate language.  It is handled by function __RegisterThreadDataSize.

ArmstrongJ:
I'll have a look at how Linux handles TLS.  Give me a day or two, but it should be relatively easy to implement.

ArmstrongJ:
The pthreads implementation does implement `pthread_key_create` for thread-specific data, basically storing a linked list in the pthread implementation's internal thread data structure.  However, I don't think it would be correct to call back into the pthread implementation itself since only pthreads would actually have the structures initialized.  I think a similar implementation in mthread would suffice.

The work I've done really doesn't rely on any registers per se.  A lot of other thread implementations on Linux do rely on using certain registers for tracking thread data, but I don't think we should have to.  It keeps things more portable if we don't.

jmalak:
OK. I think use of some register has sense for performance reason. Most OS-es have specialized functions or mechanism how to quickly access TLS.
You don't need to search TLS data in some linked list and get it by one simple instruction.
As starting point I can do search in linked list for thread TLS and later we can optimize it for performance together with pthread, I think there should be similar problem.

rdos314:
Both RDOS and Windows uses FS register to access TLS data.

jmalak:
Sorry I was not exact, I mean performance problem in pthread implementation if number of active threads will be high.

ArmstrongJ:
Most OSes do have a way to access TLS data, like `TlsGetValue` on Windows.  However, Linux is a different beast as you might have seen.  We basically only have the Linux syscalls to rely on rather than the rich known-to-be-present Windows system libraries.  Furthermore, Linux's definition of a thread is drastically different than, again, for example, Windows; a Linux thread is effectively a separate process that can share memory with some special pid handling to cover up that it isn't the same process.  So no, Linux doesn't have a defined way to handle TLS because the thread interface in Linux is drastically lower level than, for example, Windows.  That's why Linux has pthreads.

We shouldn't be worried about a performance hit by linked list searches.  I think that's an extremely premature optimization.  OpenWatcom is pretty far behind on single-thread performance on modern processors, so let's not worry about performance for TLS.  The big *performance hit* in pthreads (which hasn't been proven to be a problem yet, let's remember) will be searching the linked list of thread information for the current thread, not searching the thread's linked list of TLS data.

If you look at, for example, [musl libc](http://www.musl-libc.org/) and their thread implementation, I think you'll see that things can become very dirty quickly.  Let's keep ours simple and slow (although, again, it's premature to say that it is slow at all).

Using the linked list keeps the whole implementation processor-agnostic.  We should remain that way for a while because we simply don't have enough people to keep implementing these types of things for x86/amd64/mips/arm/whatever's next.  Let's keep it as simple as possible and worry more about a complete implementation.

@jmalak are you implementing the TLS inside bld/clib/mthread?  I'd like to try it myself if you can give me a few days.

rdos314:
On Windows, you really don't need the Windows system libraries. The library function TlsGetValue is just a wrapper for accessing things in the FS segment. Also, I think the run-time libarary might be able to setup the correct context of the FS or GS register per-thread. In RDOS, it's done in the kernel-mode loader module, but I see no reason why the runtime-library cannot do it provided FS or GS is saved in context switches.

jmalak:
OK. It is all for you. 
I agree that better is to do simpler consistent solution even if it is slower than other solutions. The performance have sense if number of threads is high, but for a few threads it is not so important.

ArmstrongJ:
@rdos314 I think you understand what I meant, though.  On Windows, I don't care about a register; I simply call `TlsGetValue` and Windows or the compiler (it makes no difference from the developer's standpoint) handles the rest.  Also, saying you could modify the FS or GS register per-thread would imply that you're on an x86 derivative, whereas Windows can target a whole host of CPUs (although, these days, mostly x86, amd64, ARM, and PowerPC (XBox)).  

Also, I'm pretty bad with such low-level coding, so I'd definitely have to bow out if we start down that path.  I'd have little to offer.  But I'm open to improvements, of course.  My way certainly isn't the generally accepted way of handling this stuff.

@jmalak I am glad you agree.  The performance hit will also be decreased substantially if I go back and properly implement thread detachment (it "works" right now, but it should be better) since it should remove threads from the linked list of thread data.

jmalak:
Please, implementing the TLS stuff inside bld/clib/mthread for Linux to be realy multi-threaded.
Inportant part is Registration of TLS space for Fortran and C++.
I think relevant function __LinuxAddThread and __LinuxRemoveThread are now blank.

jmalak:
Some idea can be get from 32-bit OS/2 code.

ArmstrongJ:
Quick update, but I'm currently looking through the documentation for using the kernel's thread local storage syscalls that should provide a way to do everything _correctly_.  Things might take a bit of time, but we should end up with a proper TLS implementation on Linux.

jmalak:
I understand, I little bit study TLS implementations on Linux. It is hard, no much info is available or are only general info. Relatively transparent code is in musl in glibc it is not transparent due to many macros in code. Next problem is implementation of TLS in compiler and appropriate ELF stuff in CRTL, but it is not necessary now to implement.

ArmstrongJ:
@jmalak do you have some example C++ or Fortran code that was failing due to the failure to initialize TLS?  I have a solution completed that I'd like to test.  Sorry for the long delay.

jmalak:
I don't know about any. There are C/C++ regression tests but only for Windows and OS/2.
It relates to DLL that is not usable for Linux.

jmalak:
now CRTL is initialized only in single-threaded mode, no multi-threaded mode.

PR #307, opened 2017-01-15
std::streamsize type introduced.

ak-ambi:
I haven't tested that (compilation takes ages), but changes are quite straightforward: I just used streamsize typedef instead of int type directly wherever C++ standard says to use std::streamsize.

Again, without this change CMake could not be built, because it was using std::streamsize type not defined by the Watcom's C++ library.


jmalak:
The build test is failing.
Please, review your change.

ak-ambi:
Done :)

jmalak:
Thanks for your contribution.
It is submited

PR #313, opened 2017-02-13
a few directx header fixes

sezero:
A few fixes to dinput.h and dsound.h, so that code can be compiled against them.


jmalak:
Thanks for your contribution.

sezero:
Sure.

PR #316, opened 2017-02-26
add HMONITOR_DECLARED macro to avoid duplicating HMONITOR typedefs

sezero:


jmalak:
Thanks for your contribution.
I did small change to be more MS compatible.
After build check I will submit it to repository.

jmalak:
submited

Issue #317, opened 2017-02-28
_tempnam() & co should accept const parameters


sezero:
At present we have _tempnam(char*,char*) (same for the wide char variant)
which should really be _tempnam(const char*,const char*)
The present form results in warnings when one passes const char* params
to the function.

jmalak:
I agree, we should change it.
There is no compatibility issue with any C standard, because it is Watcom extension.

sezero:
MSVC provides a _tempnam() and it accepts both its params as const,
and same is true for POSIX tempnam(). Therefore, changing it will make
it actually more compatible.

sezero:
The attach patch should do it (you can 'git am' it)

[0001-constify-_tempnam-and-_wtempnam.txt](https://github.com/open-watcom/open-watcom-v2/files/811146/0001-constify-_tempnam-and-_wtempnam.txt)

jmalak:
please create pull request if possible for tracking

sezero:
Done.

jmalak:
Thanks, submited.

PR #318, opened 2017-03-01
constify _tempnam() and _wtempnam()

sezero:
For issue #317


jmalak:
thanks for your contribution

PR #321, opened 2017-03-13
[CXX11]Removal of auto as storage class specifier (N2546)

bmanga:
Initial work towards supporting some C++11 features.
This change generates an error if auto is used as a storage specifier in c++11 mode (compiler flag -za0x).

### DIFF
--- a/bld/plusplus/c/type.c
+++ b/bld/plusplus/c/type.c
@@ -2442,6 +2442,10 @@ DECL_SPEC *PTypeStgClass( stg_class_t val )
     DECL_SPEC *spec;
 
     spec = makeDeclSpec();
+
+    if( CompFlags.enable_std0x && val == STG_AUTO ) {
+        CErr1 ( ERR_CXX11_AUTO_STORAGE_SPECIFIER );
+    }
     spec->stg_class = val;
     return( spec );
 }
--- a/bld/plusplus/gml/messages.gml
+++ b/bld/plusplus/gml/messages.gml
@@ -80,6 +80,13 @@
 If this message appears, please report the problem directly to the
 Open Watcom development team. See http://www.openwatcom.org/.
 
+:MSGSYM. ERR_CXX11_AUTO_STORAGE_SPECIFIER
+:MSGTXT. 'auto' is no longer a storage specifier in C++11 mode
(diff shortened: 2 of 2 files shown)


bmanga:
I am still trying to figure out the testing suite. Any help with it would be welcome.

jmalak:
what you mean by testing suite?
We use Travis CI, but only for compilation check.

bmanga:
I meant the "plustest" suite, which (I assume) should check for regressions and bugs in the c++ fontend. Is it still maintained?

bmanga:
I don't think that the #pragma warning errors that cause the travis build failure have anything to do with this change.

jmalak:
regression tests are unfortunately maintained, but is compatible with very old C++ standard.
They should be OK for native OW C++ compiler. But now there is problem with some test for 16-bit Windows target due to bug in CRTL, which I must fix.

jmalak:
The problem with error you fixed. Generaly each error message has some number and it is used in documentation etc. therefore if you add new message than you should add it on the end of existing messages list to not break any existing number.

jmalak:
one note to regression test, they are not fully working on Linux.

jmalak:
Sorry I am not right. bellow is result of regression test from my 32-bit Linux.

REGRESSION TESTS STARTED   : 2016-12-27, 00:40 UTC
REGRESSION TESTS COMPLETED : 2016-12-27, 01:21 UTC

	Fortran Compiler : Succeeded.
	C Compiler       : Succeeded.
	C++ Compiler     : Succeeded.
	WASM             : Succeeded.
	C run-time libr. : Failed!
		bld/clibtest/regress/direct	direct
		bld/clibtest/regress/heap	heap
		bld/clibtest/regress/mbyte	mbyte
		bld/clibtest/regress/misc	misc
		bld/clibtest/regress/process	process
		bld/clibtest/regress/safeio	safeio
		bld/clibtest/regress/safembyt	safembyt
		bld/clibtest/regress/startup	startup
		bld/clibtest/regress/streamio	streamio
		bld/clibtest/regress/time	time

C++ tests are OK only CRTL tests are not fixed to run on Linux.
Problem also exist with 64-bit version of C++ compiler on Linux.

PR #324, opened 2017-03-14
WIP: Analnode cleanups

bmanga:
Most of the functions in analnode.c have inconsistent and ambiguous names, as they use the same naming style for creating nodes (NodeCompareToZero -> creates a comparison to zero node) and for boolean tests (NodeBitField -> checks if node represents a bit field).

This requests attempts to give these functions better and consistent names.

### DIFF
--- a/bld/plusplus/c/analcall.c
+++ b/bld/plusplus/c/analcall.c
@@ -102,10 +102,10 @@ static PTREE makeCall(          // MAKE A CALL OR INDIRECT CALL
 
     if( direct_call ) {
         node = NodeUnaryCopy( CO_CALL_SETUP, proc );
-        node = NodeBinary( CO_CALL_EXEC, node, args );
+        node = NodeMakeBinary( CO_CALL_EXEC, node, args );
     } else {
         node = VfunSetupCall( proc );
-        node = NodeBinary( CO_CALL_EXEC_IND, node, args );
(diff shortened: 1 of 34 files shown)


bmanga:
I marked it as wip because if you approve of all the changes, I can squash the commits together. I kept them separate to easily revert a particular commit.

bmanga:
@jmalak any thoughts on this?

jmalak:
It is not good idea to masive change of function names, because it inroduces change tracing problems.
You should stay with existing names untill you have good reason to change function name.
You should always ask if gain is higher than pros as change tracibility etc.
Your first example as you give me I think it is good to change name of one of functions.
Even if now it can looks as nice now then in the future it can generate more problems.

bmanga:
What do you mean by tracing problems?

My main point is that working with those functions is always ambiguous and there is always a need to go dig into the source code, as there doesn't seem to be any documentation regarding these functions.
We could use this opportunity to have a consistent naming convention and start documenting the functions (maybe using doxygen or something like it).

jmalak:
change function name is always significant break for history tracing.
In C++ compiler is current to go very deep in history of code and you can not trace history by function name if it is broken by similar change. in this case you need to know original and new name and when it was changed.

bmanga:
alright, I will close this then.

PR #328, opened 2017-03-24
[CXX11] za0x now includes za (iso/ansi) flag

bmanga:
Using the -za0x flag now implies disabling extensions (-za).

### DIFF
--- a/bld/plusplus/c/cmdlnany.c
+++ b/bld/plusplus/c/cmdlnany.c
@@ -1092,6 +1092,11 @@ static void analyseAnyTargetOptions( OPT_STORAGE *data )
         CompFlags.no_alternative_tokens = true;
     }
     if( data->za0x ) {
+        // Make as standard compliant as we can.
+        CompFlags.extensions_enabled = false;
+        CompFlags.oldmacros_enabled = false;
+        CompFlags.unique_functions = true;
+
         CompFlags.enable_std0x = true;
     }
     if( data->zf ) {


pchapin:
I don't think this is a good idea. There is a difference between selecting which language standard you want to use and if you want extensions or not. Disabling extensions may rule out using certain libraries as well as other nice things. I don't want to choose between C++11 support and extension support.

For example, are near and far pointers considered an extension? If so this would mean I couldn't use C++11 when writing (realistic) 16 bit programs. That sounds like a problem.

Doesn't gcc distinguish between the use of extensions and selection of the standard? For example, the man page for gcc says this about the -std option: "When a base standard is specified, the compiler accepts all programs following that standard plus those using GNU extensions that do not contradict it."

bmanga:
@pchapin My argument is mainly for consistency. If the *a* in -za stands for ansi(iso) and turns extensions off, then -za0x should stand for ansi(iso) c++11 and turn them off as well.

I would add a -ze0x flag which keeps extensions on, or rethink the command line arguments for standard versioning.
What do you think?

pchapin:
@bmanga I agree that the names of the command line flags don't make much sense.  The short help for -za says, "disable extensions (i.e., accept only ISO/ANSI C)," which makes it sound like there is only one standard. There is also a -zA flag that is described as "disable all extensions (strict ISO/ANSI C)." I wasn't aware of that flag, and I don't know what it does.

Since I'm feeling a little less lazy today, I looked up the detailed description of these options in the Open Watcom documentation. For -za the macro NO_EXT_KEYS is defined. I think support for extended keywords is also turned off. This removes support for __near, __far, and various other keywords that start with two underscores.

Using -za (and turning on NO_EXT_KEYS) also hides various declarations in the header files. Just as a sample, <string.h> no longer declares functions like strtok_r and strspnp. In other words, various non-standard library functions are removed from visibility. I think this is too much to turn off if one just wants to activate C++11 features.

The -zA option does not appear to be documented.

Probably we went down the wrong path when we used -za0x to turn on C++11 features. That's where the inconsistency happened. Fundamentally the problem is that Open Watcom was never originally designed to support multiple standards.

I think the best way forward is to either:

1. Don't try to support multiple standards (activate what -za0x does by default and eliminate the flag completely).

2. Come up with a better way to describe which standard is in use rather than building off the -za flag which does something fundamentally different. I think this second option is what you are suggesting above, @bmanga.

Considering how few people are trying to use Open Watcom with C++11 features (since so few such features are actually implemented), dropping the -za0x flag, ether completely or for the sake of a new and better flag, would be relatively painless at this point.

If we do create a new flag it definitely shouldn't be -ze0x. I'd be strongly against that. For one thing the standard isn't called "C++ 0x" any more. We also want to be sure there is a clear path for supporting future standards. Finally -ze already exists (it is explicitly the opposite of -za). How about -zst? It appears unused. So we have -zst=c++98 or -zst=c++11. This syntax supports possible future application to the C compiler as well as future standard revisions.

I tend to support option #1 above (support only one standard). While creating the necessary command line flags is easy enough, keeping them properly separated in the compiler and library will require quite a bit of effort. I'm skeptical that we have the resources for that, and it will just blunt our work trying to modernize the compiler.

BTW, I see there are several command line flags mentioned in the short help that are undocumented. I suppose we should fix that at some point.

bmanga:
@pchapin I would be against merging everything inside a single standard, because at that point we would either break all the existing code relying on the current status of the compiler (we already have with *auto*), or never be able to properly implement c++11 (a list of breaking changes in [this so post](http://stackoverflow.com/questions/6399615/what-breaking-changes-are-introduced-in-c11) ).

Yes, your option 2 is what I suggested as well, and I think at this point it would be the best compromise.
I will post an issue regarding this matter so other people can provide their views, and close this pull request.

Issue #340, opened 2017-06-05
Nonsensical code in fold.c


bmanga:
I am not sure who to ask, but I have encountered some code that I think is clearly misguided:
https://github.com/open-watcom/open-watcom-v2/blob/a745276efa9d5c7cdb1013a11b48c2c9083bae51/bld/plusplus/c/fold.c#L1172-L1185

``notFoldable`` checks that the expression is not a zero constant, and yet inside the ``if`` the code tries to check if either is zero and to ``makeTrueFalse`` if that's the case.

I think either that part of the code is in the wrong if statement, the if condition is wrong, or the implementation of ``notFoldable`` is wrong. I am not sure which it is.

Any ideas?

bmanga:
based on the rest of the code in fold.c I would assume that it should have been : 
``if( !notFoldable ) .... ``. 
I will create a pull request, but feel free to show me wrong.

jmalak:
I don't see a problem there. 
Generaly nonFoldable should be true if argument is not constant expression, I thing it is OK in existing code. If both operand are non-foldable then it returns without any action.
If one operand is foldable then it is possible to do some optimization. Function should be symetric, the same for each argument if it is foldable.


Anyway line 1171 is wrong, it should be:
has_decoration = has_decoration_left | has_decoration_right;
Nothing danger, it is used only for debugging.

bmanga:
@jmalak that's exactly my point. If execution reaches inside the if statement, then neither ``op1`` nor ``op2`` can be constant expressions.
For example, [this line](https://github.com/open-watcom/open-watcom-v2/blob/a745276efa9d5c7cdb1013a11b48c2c9083bae51/bld/plusplus/c/fold.c#L1180) can never evaluate to true.

I am trying to clean up the file, I will create a pull request sometime soon and we can analyze the code from there.

jmalak:
Sorry I didn't study code in details. Yes, this code looks like dead code.

bmanga:
@jmalak does OW provide any backend code optimization passes?
That is, is it necessary to apply the folding logic on the AST in fold.c or could that be done at a later time?
As it is fold.c seems to only perform some optimization if the lhs is constant:
``1 && X -> X``
but not
``X && 1 -> X``.
I am wondering if those are dealt with at a later phase or it would be useful to add me as part of the refactoring.

jmalak:
I think OW compilers front ends are single pass. It does only simple transparent optimisation.
[...]

PR #341, opened 2017-06-05
Foldfix

bmanga:
Reworked some of the code. Now some simple transformations are correctly applied.
Although I have tried some basic examples and they work correctly, this should require some more careful testing to make sure nothing broke.

### DIFF
--- a/bld/plusplus/c/fold.c
+++ b/bld/plusplus/c/fold.c
@@ -502,8 +502,6 @@ static PTREE castConstant( PTREE expr, TYPE type, bool *happened )
     return( expr );
 }
 
-#if 0
-
 static bool soFarSoGood( PTREE expr, unsigned op, CGOP cgop )
 {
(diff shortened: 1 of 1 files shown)


bmanga:
This closes issue #340 .

bmanga:
I couldn't find any regressions for this patch.

jmalak:
Give me enough time, because it is very hard to follow your changes.
It is as new code, that I will need to study it in more details.
Anyway, better is to modify existing code by minimal way until it is necessary to do massive code changes.

bmanga:
I rely on the constant folding to actually work for a following pull request implementing nullptr conversions to bool, that's why I had to fix this.

jmalak:
Submitted as squashed commit bb4a67aac3a78f7e5392b85e4944d8b8fb156219
Thanks for your contribution

PR #360, opened 2017-07-29
nt/GL/glu.h: fix gl.h include (uppercase 'GL' directory)

sezero:


jmalak:
Thanks for your contribution

Issue #357, opened 2017-07-24
Broken C++ compiler


jmalak:
I run C++ compiler regression tests and found out some problems with constant expressions.
The failing tests are diagnostics diag0155.c and positive float03.c.
@bmanga Please, could you review these issues.

bmanga:
I will have a look tomorrow.

bmanga:
I could not find a .chk file for float03.cpp, but the regressions on diag0155.c seem solved.
Could you please let me know if #362 fixed this?

jmalak:
It looks like some mistake in code. Travis build reports bug.

jmalak:
float03 is real program that must be succesfuly run.
It reports problem, output should be in bld\plustest\regress\positive\*.sav files

bmanga:
I think it's fixed now.

jmalak:
Now build is not failing, but more regression tests are failing.
Positive group failed float03, opt03, new01 and diagnostics tests are failing
You must test it for 32-bit and 16-bit Intel architecture.
You can run C++ regression tests from bld/plustest by command "builder buildtest" (full set) or from bld/plustest/regress/apropriate test sub-directory by command "wmake"

jmalak:
I look into it a little bit and diagnostics looks OK.
It was necessary to update reference output file for 32-bit and 16-bit architectures.
I have some notes to diagnostics messages for diag0157.c
source\diag0157.c
(3,12): Warning! W690: conditional expression is always false (zero)
(27,9): Warning! W367: conditional expression in if statement is always true
(30,5): Warning! W013: unreachable code
(34,9): Warning! W690: conditional expression is always false (zero)
(34,9): Warning! W368: conditional expression in if statement is always false
(35,9): Warning! W013: unreachable code
On the line 34 is problem diagnosed 2 times with two different messages.
We should use always only one of following message
WARN_IF_ALWAYS_TRUE = 367
WARN_ALWAYS_TRUE = 689
WARN_IF_ALWAYS_FALSE = 368
WARN_ALWAYS_FALSE = 690
I think message on line 34 should be WARN_IF_ALWAYS_FALSE
On the line 37 is missing diagnostics message, probably should be WARN_ALWAYS_TRUE.
Sorry, I am not C++ expert, but is guarantied that "this" is always non-zero?

bmanga:
That is due to the fact that open watcom does boolean checks all over the place.
[...]

PR #362, opened 2017-07-30
Make folding more aggressive, fix a bug

bmanga:
Should address issue #357


jmalak:
Thanks for fix. 
I will run full regression test to check if all is OK.

PR #370, opened 2017-08-12
[CXX11] Implement stateless lambdas

bmanga:
Added initial implementation of stateless lambdas (without lambda captures).
Notes:
* These lambdas require trailing return types, as the compiler is unable to perform type deduction.
* Captures are not implemented
* Lambda declarator is mandatory (no `[]{return 0}`)
* They are not allowed in any context other than block / function scope
* They are default constructible / copiable, unlike what the standard mandates (but they may become so in c++20)


bmanga:
@jmalak do you know why one of the generated ytab.c files is not built properly? 
the one in `bld/plusplus/386/nt386.rtd` does not contain the `#include "lambda.h" which I added to the `yydriver.c` file.

jmalak:
One note, each new source file should have header

```
/****************************************************************************
*
*                            Open Watcom Project
*
* Copyright (c) 2017-2017 The Open Watcom Contributors. All Rights Reserved.
*
*  ========================================================================
*
*    This file contains Original Code and/or Modifications of Original
*    Code as defined in and that are subject to the Sybase Open Watcom
*    Public License version 1.0 (the 'License'). You may not use this file
*    except in compliance with the License. BY USING THIS FILE YOU AGREE TO
*    ALL TERMS AND CONDITIONS OF THE LICENSE. A copy of the License is
*    provided with the Original Code and Modifications, and is also
*    available at www.sybase.com/developer/opensource.
*
*    The Original Code and all software distributed under the License are
*    distributed on an 'AS IS' basis, WITHOUT WARRANTY OF ANY KIND, EITHER
*    EXPRESS OR IMPLIED, AND SYBASE AND ALL CONTRIBUTORS HEREBY DISCLAIM
*    ALL SUCH WARRANTIES, INCLUDING WITHOUT LIMITATION, ANY WARRANTIES OF
*    MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, QUIET ENJOYMENT OR
*    NON-INFRINGEMENT. Please see the License for the specific language
*    governing rights and limitations under the License.
*
*  ========================================================================
*
* Description:  WHEN YOU FIGURE OUT WHAT THIS FILE DOES, PLEASE
*               DESCRIBE IT HERE!
*
****************************************************************************/
```

I forgot next comment for lambda.h.
It should looks like sample below, because it is used for parser only that necessary header files is certainly included.

```
#ifndef _LAMBDA_H
#define _LAMBDA_H

extern void       LambdaStartClosure(PARSE_STACK *state);
extern PTREE      LambdaFinishClosure();
extern DECL_INFO *LambdaMakeClosureCallOpDeclaration(DECL_INFO *parameter_declaration_clause, DECL_SPEC *trailing_return_type, TOKEN_LOCN *locn);
extern void       LambdaAttachBodyToCallOp(DECL_INFO *lambda_function);

#endif

```

I don't have a problem with missing lambda.h reference in ytab.h.
It looks like your ytab.c was not updated for some reason.

bmanga:
I don't like relying on the header files being transitively included. That caused me all sorts of problems during development because I was including files that do not include their dependencies at the start and I had to manually go figure out which headers I had to include.
If a header is already included, it will just be `#ifndef`ed out, but if it's not, it will work as expected.

bmanga:
@jmalak Does this look okay now?

bmanga:
ping

jmalak:
Change looks OK.
Now I am working on fixing C++ compiler after previous changes which break some compiler features. As soon as I will finish and C++ compiler tests will pass I will apply this change.

Issue #380, opened 2017-10-02
download latest stable pre-compied open-watcom


sagudev:
Where can I download latest stable open-watcom installer? (for Windows)

- Sourceforge has got installers from 2015
- http://www.openwatcom.org/download.php has version 1.90
There are no latest pre-compied open-watcom for other OS too.

sezero:
At present, there are no newer official releases than 1.9.0.
Your options are:

- snapshot builds from the official OW perforce repository, includes installers:
  https://efbe.musca.uberspace.de/

- snapshot builds from this fork of OW (no installers):
  https://github.com/open-watcom/travis-ci-ow-builds/archive/master.zip
  https://github.com/open-watcom/travis-ci-ow-builds

sagudev:
maybe it will be good to add 

> snapshot builds from the official OW perforce repository, includes installers:
> https://efbe.musca.uberspace.de/
to README.md file so evryone could find latest installers.

Maybe it will be good to make some project releases and/or tags.

- [ ] add https://efbe.musca.uberspace.de/ to README.md
- [ ] make some project releases and/or tags

jmalak:
First, this project is fork of OW 1.9 to port OW to 64-bit, and do next development, because official development is dead. 
Therefore installers from OW 1.9 have no sense for this project.
We can regularly create installers but only for 32-bit version of OW.
64-bit version you must install manualy,
Anyway manual installation is simple see [readme](https://github.com/open-watcom/travis-ci-ow-builds)

ArmstrongJ:
Just a reminder that nightly builds in simple zip archives are also available at:

http://buildbot.approximatrix.com/releases/

sagudev:
Ok I understand OW 1.9 is not this github project, but still why the project doesn't have download links for installers at least in the [main site](http://open-watcom.github.io/open-watcom/)

XVilka:
To be honest would be nice to have git tag for sources as well, for packaging in Linux/BSD distributions.

jmalak:
@sagudev  the reason is my limited free time.
Installers has low priority on my to do list, because I am concentrated now to more important C++ compiler and it's regression tests to fix some problem with 64-bit version of C++ compiler.
If you are interested in creation of installers regular way in Travis build then you can help.
All the necessary staff exist in OW, only need to migrate appropriate commands from OW build server to script for new stage in Travis build process and ensure back transfer of build installers to GitHub repository.
[...]

PR #381, opened 2017-10-02
Update README.md

sagudev:
add download section

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,15 @@ Open Watcom v2 Fork
 This document gives a quick description of the source tree layout. For more
 information about the project and build instructions see the GitHub wiki.
 
+Download
+------------------
+Snapshot builds from the official OW perforce repository, includes installers:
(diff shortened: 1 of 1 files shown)


sagudev:
See #380

jmalak:
I don't think it is acceptable change, because it is unrelated to our project.
Main changes between OW 2.0 fork and official OW 1.9 version is that we have 64-bit version of OW toolchain and many enhancement which doesn't exist in OW 1.9.
We ensure maximum compatibility with OW 1.9 if possible, but not all is the same.
OW 2.0 has many features which doesn't exist in OW 1.9, by example support for PE+ format and all tools are native 64-bit applications. We also implemented POSIX threads functionality and some new standard feature for C++. We improved compatibility with ISO and POSIX standards against OW 1.9.
Official OW 1.9 is in reality only subset of OW 2.0 fork functionality.

Issue #356, opened 2017-07-17
bad node in heap on DOS for minimal program C & C++


pgrAm:
Hi,

I think I have found a bug in the C runtime library for DOS. I have a simple program which just checks the heap without explicitly allocating anything, with the most recent version of OW for DOS this finds a bad node in the heap, the same program compiled with OW for Windows or DOS/4GW results in no such bad node.

here is the minimal program (main.c):

```
#include <malloc.h>
#include <stdio.h>

void heap_dump()
{
	
	struct _heapinfo h_info;
	int heap_status;
	
	FILE* log = fopen("heap.txt", "w");
	
	size_t heapSize = 0;
	
	h_info._pentry = NULL;
	for (;;) {
			heap_status = _heapwalk(&h_info);
			if (heap_status != _HEAPOK) break;
			fprintf(log, "  %s block at %p of size %4.4X\n",
					(h_info._useflag == _USEDENTRY ? "USED" : "FREE"),
					h_info._pentry, h_info._size);
	
			heapSize += h_info._size / 1024;
	}
	
	fprintf(log, " heapsize: %u KBytes\n", heapSize);
	
	switch (heap_status) {
	case _HEAPEND:
			fputs("OK - end of heap\n", log);
			break;
	case _HEAPEMPTY:
			fputs("OK - heap is empty\n", log);
			break;
	case _HEAPBADBEGIN:
			fputs("ERROR - heap is damaged\n", log);
			break;
	case _HEAPBADPTR:
			fputs("ERROR - bad pointer to heap\n", log);
			break;
	case _HEAPBADNODE:
			fputs("ERROR - bad node in heap\n", log);
	}
	
	fclose(log);
}

int main(int argc, char** argv)
{
	heap_dump();
}
```

the output of the program for 16bit DOS
```
  USED block at 15a1:0016 of size 0014
  USED block at 15a1:002a of size 0202
  FREE block at 15a1:022c of size 1DD0
 heapsize: 7 KBytes
ERROR - bad node in heap

```

and on Windows:

```
  USED block at 001d002c of size 002B
  USED block at 001d0084 of size 002B
  USED block at 001d0094 of size 002B
  USED block at 001d00ac of size 002B
  USED block at 001d00bc of size 002B
  USED block at 001d0ac4 of size 002B
  USED block at 001d0bdc of size 002B
  USED block at 001d0c04 of size 002B
  USED block at 001d0c2c of size 002B
[...]

PR #384, opened 2017-10-18
fix heapwalk

udovichenkoAlexnder:
fix for issue https://github.com/open-watcom/open-watcom-v2/issues/356
after [this refactoring](https://github.com/open-watcom/open-watcom-v2/commit/fdfd118c51044a854e39473818857205968f5525#diff-4e91dde9823d0cc23f586c06080d4b45L133) comparison is done with Ien field instead of pointer itself


jmalak:
Thanks for your contribution.
Just submitted.

PR #394, opened 2017-12-02
Added few missing std::vector and std::set functions and const find() and count() for RedBlack tree

Wohlstand:
Those functions are in standard and there are missing here

http://en.cppreference.com/w/cpp/container/vector/data
http://en.cppreference.com/w/cpp/container/set/count


jmalak:
Thanks for your contributions

PR #399, opened 2018-01-29
Added .count() method to std::map

pgrAm:
In accordance with the standard I added a .count method for OW's std::map implementation


jmalak:
Thanks for your contribution.
submitted

Issue #284, opened 2016-10-21
building on osx fails


jeroenjacobs79:
Attempting to build the Watcom source tree on the latest OSX Sierra.

Got an error here:

```
=== 16:03:47 /Users/jeroenjacobs/open-watcom-v2/bld/browser/dlgprs/binbuild ===
yacc dlgytab.h
yacc bndytab.h
c++ bind.obj
In file included from ../cpp/bind.cpp:39:
In file included from ../h/chbffile.h:40:
In file included from ../h/chfile.h:37:
/Users/jeroenjacobs/open-watcom-v2/bld/watcom/h/wstd.h:144:6: error: OS_xxx macro not enabled.
    #error OS_xxx macro not enabled.
     ^
1 error generated.
c++ dialog.obj
In file included from ../cpp/dialog.cpp:36:
In file included from ../h/dialog.h:40:
/Users/jeroenjacobs/open-watcom-v2/bld/watcom/h/wstd.h:144:6: error: OS_xxx macro not enabled.
    #error OS_xxx macro not enabled.
     ^
In file included from ../cpp/dialog.cpp:36:
../h/dialog.h:66:41: warning: private field '_name' is not used [-Wunused-private-field]
            const char *                _name;
                                        ^
1 warning and 1 error generated.
c++ main.obj
In file included from ../cpp/main.cpp:35:
In file included from ../h/chfile.h:37:
/Users/jeroenjacobs/open-watcom-v2/bld/watcom/h/wstd.h:144:6: error: OS_xxx macro not enabled.
    #error OS_xxx macro not enabled.
     ^
1 error generated.
c++ scanner.obj
In file included from ../cpp/scanner.cpp:39:
In file included from ../h/chbffile.h:40:
In file included from ../h/chfile.h:37:
/Users/jeroenjacobs/open-watcom-v2/bld/watcom/h/wstd.h:144:6: error: OS_xxx macro not enabled.
    #error OS_xxx macro not enabled.
     ^
1 error generated.
c++ styles.obj
In file included from ../cpp/styles.cpp:43:
In file included from ../h/styles.h:35:
/Users/jeroenjacobs/open-watcom-v2/bld/watcom/h/wstd.h:144:6: error: OS_xxx macro not enabled.
    #error OS_xxx macro not enabled.
     ^
1 error generated.
c++ bndytab.obj
In file included from bndytab.cpp:5:
In file included from ../h/bind.h:40:
[...]

PR #414, opened 2018-03-25
Fix a buffer overflow in rc

andersca:
Increase the debug string length so it can hold the maximum signed long value.

When lexing certain numbers (such as WS_POPUP which is #defined to 0x80000000L), the resulting string gets too big and overflows the debugstring buffer.

Fix this by making the array 21 characters long, which is enough to represent the longest possible 64-bit number (-9223372036854775808), including the null character.

(This fixes the rc crash mentioned in https://github.com/open-watcom/open-watcom-v2/issues/284)


jmalak:
Many thanks for your contribution

PR #465, opened 2018-07-22
Fixed bug where dosbox would fail to run due to path not being expanded.

samizzo:


samizzo:
Sorry I should have created a branch for this fix!

PR #467, opened 2018-07-22
Fixed bug where dosbox would fail to run due to path not being expanded.

samizzo:
dosbox was failing on macOS (and probably haiku and bsd) unless the dosbox binary was in the current path. This was due to the macro not being expanded in the build file.


samizzo:
Ah, this should be applied after PR #466.

PR #488, opened 2018-11-06
bld/wasm/c/direct.c: missing proto for func heap

ideafarm:
defined only if debug flag set


jmalak:
Thanks for fix.
Implemented as 7d73c1037e1c0660a18f7623a5b299db4eb7e5db comit.

jmalak:
Thanks for fix.
Implemented as 7d73c1037e1c0660a18f7623a5b299db4eb7e5db comit.

PR #490, opened 2018-11-08
comment out debugging code in wv that won't compile

ideafarm:
The code that is now commented out unconditionally was previously commented out unless debug=1.  It won't compile.  With these changes, wv can be built using "builder build OWDEBUGBUILD=1".

Testing:  I can now run wdw wdw [program] and step into debugger code.
Regression testing:  After building with debug=0, I released to my IPDOS (tm) craftwork environment and used it to do a full build of IPDOS and to then test that build by running it in production.

### DIFF
--- a/bld/wv/c/dbgwintr.c
+++ b/bld/wv/c/dbgwintr.c
@@ -31,6 +31,10 @@
 
 
 #ifndef NDEBUG
+
+#if defined( NEVERdEFINED )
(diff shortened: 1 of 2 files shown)


jmalak:
It is unacceptable to remove any code for reason that it want compile.
It is necessary fix it.
Right fix is submited by commit 74c79761ecb6ae87d0f66ecc76bd9c083e0baaf1
Anyway thanks for bug report.

PR #491, opened 2018-11-09
Add missing or update file description

ideafarm:
I'm going to take a crack at debugging the debugger. It's a mess. For example, the Locals window is totally empty when running wdw on wdw. (When running wdw on IPDOS executables, the window lists the local variables, although the values of many are "????".) As I begin to study the code, I will add file comments for files that don't have anything at all and that are linked into wdw.

In adding comments, I'm being conservative. (It's better to say nothing than to add a comment that is inaccurate.)

### DIFF
--- a/bld/gui/c/guifcrck.c
+++ b/bld/gui/c/guifcrck.c
@@ -2,6 +2,7 @@
 *
 *                            Open Watcom Project
 *
+* Copyright (c) 2002-2018 The Open Watcom Contributors. All Rights Reserved.
 *    Portions Copyright (c) 1983-2002 Sybase, Inc. All Rights Reserved.
 *
 *  ========================================================================
@@ -24,8 +25,10 @@
 *
 *  ========================================================================
 *
-* Description:  WHEN YOU FIGURE OUT WHAT THIS FILE DOES, PLEASE
-*               DESCRIBE IT HERE!
+* Description:  Not meaningful.  Called early by some process executives,
+*               perhaps to force module inclusion by linker.
+*
+*               Example: \watcom\binnt\wdw.exe (wv built for Windows NT)
 *
 ****************************************************************************/
 
--- a/bld/gui/c/guifcrck.c
+++ b/bld/gui/c/guifcrck.c
@@ -28,7 +28,9 @@
 * Description:  Not meaningful.  Called early by some process executives,
 *               perhaps to force module inclusion by linker.
 *
-*               Example: \watcom\binnt\wdw.exe (wv built for Windows NT)
+*               This module is known to be linked into the following
+*               executable files:  \watcom\binnt\wdw.exe (wv built for
+*               Windows NT)
 *
 ****************************************************************************/
 
--- a/bld/gui/c/guiutil.c
+++ b/bld/gui/c/guiutil.c
@@ -2,6 +2,7 @@
 *
 *                            Open Watcom Project
 *
+* Copyright (c) 2002-2018 The Open Watcom Contributors. All Rights Reserved.
 *    Portions Copyright (c) 1983-2002 Sybase, Inc. All Rights Reserved.
 *
 *  ========================================================================
@@ -24,8 +25,10 @@
 *
 *  ========================================================================
 *
-* Description:  WHEN YOU FIGURE OUT WHAT THIS FILE DOES, PLEASE
-*               DESCRIBE IT HERE!
+* Description:  Utility functions: string manipulation, process arguments,
+*               root window.
+*
+*               Example: \watcom\binnt\wdw.exe (wv built for Windows NT)
 *
 ****************************************************************************/
 
--- a/bld/gui/c/guiutil.c
+++ b/bld/gui/c/guiutil.c
@@ -28,7 +28,9 @@
 * Description:  Utility functions: string manipulation, process arguments,
 *               root window.
 *
-*               Example: \watcom\binnt\wdw.exe (wv built for Windows NT)
+*               This module is known to be linked into the following
+*               executable files:  \watcom\binnt\wdw.exe (wv built for
+*               Windows NT)
 *
 ****************************************************************************/
 
--- a/bld/gui/win/c/guildstr.c
+++ b/bld/gui/win/c/guildstr.c
@@ -2,6 +2,7 @@
 *
 *                            Open Watcom Project
 *
+* Copyright (c) 2002-2018 The Open Watcom Contributors. All Rights Reserved.
 *    Portions Copyright (c) 1983-2002 Sybase, Inc. All Rights Reserved.
 *
 *  ========================================================================
@@ -24,8 +25,9 @@
 *
 *  ========================================================================
 *
-* Description:  WHEN YOU FIGURE OUT WHAT THIS FILE DOES, PLEASE
-*               DESCRIBE IT HERE!
+* Description:  Utility functions: string loading from resource file.
+*
+*               Example: \watcom\binnt\wdw.exe (wv built for Windows NT)
 *
 ****************************************************************************/
 
--- a/bld/gui/win/c/guildstr.c
+++ b/bld/gui/win/c/guildstr.c
@@ -27,7 +27,9 @@
 *
 * Description:  Utility functions: string loading from resource file.
 *
-*               Example: \watcom\binnt\wdw.exe (wv built for Windows NT)
+*               This module is known to be linked into the following
+*               executable files:  \watcom\binnt\wdw.exe (wv built for
+*               Windows NT)
 *
 ****************************************************************************/
 
--- a/bld/gui/win/c/guimenus.c
+++ b/bld/gui/win/c/guimenus.c
@@ -2,7 +2,7 @@
 *
 *                            Open Watcom Project
 *
-* Copyright (c) 2002-2017 The Open Watcom Contributors. All Rights Reserved.
+* Copyright (c) 2002-2018 The Open Watcom Contributors. All Rights Reserved.
 *    Portions Copyright (c) 1983-2002 Sybase, Inc. All Rights Reserved.
 *
 *  ========================================================================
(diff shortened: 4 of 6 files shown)


jmalak:
Thing about code on project level.
Debugger uses AUI (GUI + WPI + COMMONUI + UI), DIP (DIG + DWARF), MAD (DIG), TRAP (DIG + TOOLS) projects.

ideafarm:
I agree that my file comments can be made clearer, and will edit them to address some of your concerns and make another pull request.  Since I don't agree with some of your objections, and you are the gatekeeper, you will have to decide whether a comment that you don't like, but is accurate, is worse than no comment at all.

ideafarm:
P.S., the pull title is a typo; it should be WV.  I suspect that wv is being built with line numbers only rather than with all symbols, and will look into that.  Thanks for the rapid responses and for the corrected build fix for wv.  It's going to take me some time to get up to speed with this code; so far it looks pretty bad (no comments, nearly meaningless variable and function names, random placement of functions in modules, duplication and pseudoduplication of functions across modules).  I've seen worse, but not much worse.

ideafarm:
I've edited my proposed file comments.  They now appear to be part of this pull request, so I guess that means that they are available to you for review and acceptance.

jmalak:
I am open to any discussion, but take into acount real purpose of file description.
Check existing files and you will see what rules are used for it.
I think there is not reason to rework it because we don't gain anything useful by it.
Anyway consult [Developer guide](https://github.com/open-watcom/travis-ci-ow-builds/blob/master/docs/devguide.pdf) in OW documentation.
If you think that it should be changed than open discussion by proposing what should be changed.
If it will be valuable I think it will be accepted.
But personaly I think this is not such case.

rdos314:
Perhaps if would be a bit more specific about what problems you have with the debugger? I use it all the time on Windows 8 and 10, and I don't have any serious issues with it. I both debug Windows applications and RDOS applications with the tcp trap. I've also built the debugger on RDOS,  and that works too.

ideafarm:
Here is an UNVERIFIED list of the bugs that give me grief when using WDW.EXE on Windows 10:

(1) Module List window:  The entire list appears twice.  Clicking in the first set results in"Source not found." (approx).  This seems to be a recently introduced bug.

(2) Locals and Watch windows:  The value of many local variables is "????".  I work around this occasionally by prepending "static" to the variable allocation statement.  This bug seems to be several years old.

(3) Default radix:  The debugger becomes unstable if I try to change default behavior to radix 16.  I submitted code fixes for this years ago, which I think Mr. Malak put in.  About a year ago, the bug seemed to still be there.  I haven't tried changing radix lately.  My workaround is to change the radix for each watch variable as I need it, which is a PITA.

rdos314:
There is an issue witb dll symbol files. I avoid this issue by copying dll symbol files to the directory of the main executable,  which fixes that issue. That issue is related to not finding the source file. 

I've not seen problems with variables not being accesible, but then I usually don't write complex C code or use DOS applications. I think you need to find some pattern in when this happens. 

I don't change radix so I don't know about that.

ideafarm:
Apologies for using the word "stupid".  Should have used "not optimal".  Comment clutter is also not optimal, so I understand Mr. Malak's resistance to expanding the scope of the file comment section.  Regarding this pull request, if it is accepted, I will continue to add similar comments as I encounter more files with no file comments at all.  If it is rejected, then I just won't bother to add comments.  I recommend that this pull request be granted so that I will continue to add comments, which will capture the benefit of my newbie experiences.  Once I am familiar with the code, I will no longer be able to see just how impenetrably unreadable this code is and will no longer have any motivation to make things easier for the next newbie.

rdos314:
The two worst things about software projects is #ifdef clutter and  comment clutter. I think OpenWatcom is a rather good project in relation to those problems.

rdos314:
We have a large commercial C++ application, but it doesn't use a lot of the fancy C++ functions. It runs under RDOS.

ideafarm:
Thanks.  IPDOS (tm) uses almost none of the library functions provided by OW.  It "rolls its own" for almost everything, using OS calls.  It was originally written for 16 bit OS/2, then ported to 32 bit OS/2.  For a while, it supported both OS/2 and Windows NT. But for the past 20 years, I've focused on Windows.  So it's a good testbed for OW's code generation and debugger, but it doesn't exercise any of the libraries and it only tests using Windows 10 as the build and target.

@rdos314:  What do you use to debug?

jmalak:
Below is example for bld/gui/c/guifcrck.c source file
```
* Description:  Not meaningful.  Called early by some process executives,
*               perhaps to force module inclusion by linker.
*
*               This module is known to be linked into the following
*               executable files:  \watcom\binnt\wdw.exe (wv built for
*               Windows NT)
*
****************************************************************************/
```
What value it has?
Nothing about file content.
What has source code with linking proces?

At all no value.

jmalak:
Sorry, it is not my idea, but this is way how it is done in all OW source base. 
It looks like you don't understand my explanation or don't want understand me.
If you want to add something outside these rules you can, it is not denied.
Add it on the begining ( if it is common )  or into source file in the appropriate place.
But file description should follow rules which I described and is common for all OW file descriptions.
Code should be portable as much as possible that platform specific comment is always exception and should have sense.
Take into account that any programmer can decide that your comment is useless and remove it later.
You should think about it little more to add something what has value for all.

Note: don't add any comment about linking and compilation if no specific setup is necessary for it.
Linking and compilation is control by builder control files (*.ctl) and wmake makefiles.
These are right place to add any comment related to linking and compilation project if necessary.

Source file should be clear and hold comment related to source code in the file.

Anyway each open source project has some rules and is necessary to follow them, even if what I think about it. It has nothing with art.

rdos314:
> @rdos314: What do you use to debug?

It varies. I mainly run the graphical debugger (wdw) on Windows 7 with the tcp trap. In the target, I usually use my own server that is part of RDOS and has features like allowing multiple instances and hooking onto an already running process. I've also integrated the debug server into our loader application. The target applications I most often debug are 32-bit flat, but I also use it to debug 32-bit segmented device drivers in the kernel. I've also built & used the debugger for RDOS (character mode only), mostly to debug the debug server. Sometimes I debug local Windows programs. I've run wdw on Windows 10 as well, but not as much as on Windows 7.

ideafarm:
@ rdos314:  Mr. Ekblad, that's helpful; it gives me an idea of why and how you are using OW.  I'm currently crafting a search tool so that I can easily search the bld/ for text.  When that is ready, I will begin to study the wv/ code.

@jmalak:  Mr. Malak, this pull request was intended to address the fact that there are two definitions for GUIXMain and it does not appear to be a situation involving overload semantics.  Rather than try to clarify things by adding file comments, WDYT about me renaming the two definitions as GUIXMain1 and GUIXMain2?  That would make it easy to identify the code that uses each, which could be the first step to eliminating the redundancy.

jmalak:
There is only one definition of GUIXMain in each GUI library.
Take into account that GUI library has two version.
First is UI version (character mode) and second one is GUI version (graphical mode).
Therefore one GUIXMain is in gui/ui/c and second in GUI counterpart gui/win/c.
Character mode has two version one for 16 bit Windows (which has no character mode) and second is for OSes with character mode.
Windowed version, one is for Windows 16/32/64-bit and second is for 16/32-bit OS/2 Presentation Manager.
There was very confusing guixsys.c file which was merged with guixwind.c in gui/ui and now is counter part of guixwind.c in gui/win/c.
There is absolutely no reason for renaming GUIXMain function.

jmalak:
in gui/doc is documentation of GUI project and API

ideafarm:
No reason for renaming?  How about this one:  You just told me that there are two versions of GUIXMain because the GUI library has two versions, one for graphical mode and one for character mode.  A blind man could see the illogic of that.  "GUI" stands for "graphical user interface".

This kind of thing doesn't impede you because you've lived in the code long enough to become familiar with it.  But illogic like this will totally frustrate anyone new to the code.

It would be very helpful to me to know what your development focus is right now so that I can understand the context, and perhaps the cause, of your resistance to renaming functions, which is a very safe procedure if done competently using global verification searches.  You seem to be doing lots of cosmetic code changes, so I thought that similar changes from me would be welcome.

jmalak:
Because both versions don't have always same prototype for GUI project internal functions.
GUIXMain is GUI internal function.
Are you sure that GUIXMain prototype should be same for both versions?

ideafarm:
This isn't a priority for me.  I just wanted to get your thoughts on me taking time to submit proposed function name changes.  But what you just told me suggests that the current function names are so illogical (functions whose name contains "GUI" being used by character mode applications) that now I'm inclined to forget the whole idea of cleaning up this code as I study it.

I'm currently finishing up a tool that will enable me to search a directory tree for plain text.  When I have that ability, I'll begin to study the code and begin to debug the worst bugs in the debugger.

jmalak:
As I wrote you before the debugger uses many OW projects (each of them is relatively complex and requires good knowledge of code)
All will have the same problem for you "ilogical design for newcommer".

GUI
UI
COMMONUI
WPI
AUI
DIG
TRAP
MAD
DIP
DWARF

ideafarm:
"...does have a [GRAPHICAL user interface], but it is made with characters instead of GRAPHICS"

Your statement contradicts itself.     :)

ideafarm:
This is all in a friendly spirit; I'm not being critical.  It's a good example of what happens when coders have weak English language verbal skills, either because English is not their first language or because verbal aptitude is low.  People are not attracted to software craftwork because they like English composition and are good at it.

The problem is compounded by rapid change in this industry, which burns people out, resulting in most coders being too young to have personal experience with the appearance of various terms such as "GUI".  "GUI" came into use when individual pixels became addressable in the early 1990's and referred to such displays.  Full screen character mode displays had been in use for decades.  IBM's CICS system is perhaps the preeminent example.

Yes, I think that OW code sucks.  Do I think that it should be edited?  No.  Each change risks introducing a bug, so I think that code changes to a mature code asset like OW should be minimized.  Perhaps in the future we can contemplate a thrust in the direction of renaming functions and adding comments to make the code more readable.  But at this moment, my priority is to debug so that the debugger works perfectly again.

PR #495, opened 2018-11-23
Fix debug  build of as

ideafarm:
With debug, three arrays were constructed but never referenced.  Now they are referenced.


jmalak:
it was fixed by 0c1132e9ae31126a5a678733562cea8e4f4725e5 commit
there was debug macros mismatch and theirs incorrect use

PR #533, opened 2019-04-18
Fix issue with trap file loader used in dos version of wd

wlabelle:
The DIGLoader( Read ) function is already checking the size of
the read bytes and returning true or false. Comparing the return
value with imagesize will fail even if the file was loaded correctly.


jmalak:
Thanks for your contribution. 
It is commited now.

PR #540, opened 2019-06-01
Fix incorrect definition if LPCPROPSHEETPAGEW

markisch:
for 0x0600 > _WIN32_WINNT >= 0x0501


markisch:
Obviously you did this yourself already... Thanks, but...

PR #547, opened 2019-08-06
Include the *.cnv files in the wipfc rel target

wlabelle:
The *.cnv files used by wipfc were not being copied to rel directory by the rel target.


jmalak:
thanks for your contribution

PR #550, opened 2019-08-15
Add missing dereference for a end of string test in hlpview

wlabelle:


jmalak:
thanks for your contribution

PR #558, opened 2019-09-25
Fixes a bug where PATH variables were not quoted.

friedkiwi:
PATH variables weren't quoted properly, which caused the build scripts to fail when e.g. PATH contains a space. This typically does not occur on Linux, however, WSL tends to export Windows environment variables as well, introducing spaces.

This PR fixes this issue.


jmalak:
Thanks for your contribution, commited.

PR #587, opened 2020-02-27
fresh (Jiri's changes to getimpline()

ideafarm:


jmalak:
I don't understand why new macros are added if right macros already exists.
Why it is added to makeinit, purpose of this make file is to setup global things.
Right make file is local.mif where external compilers configuration is located.
Compiler options can be changed only for very good reason, but it is not this case.
Anyway problem is already fixed.

ideafarm:
This looks like a pull request that I created by mistake a few days ago.  Am closing.  I just completed the compiler optimization changes and have will create a pull request as soon as I can figure out how.

PR #588, opened 2020-03-01
Use VISUALC compiler optimization flags -Ob2ity rather than -Oait -Ox -O2

ideafarm:
Microsoft has deprecated -Oa and -Og.  Within the last few days, I posted source code for a simple program for which -Og produces incorrect code.  (Practically no code is emitted for a loop.)  That flag produces multiple build errors, only one of which was fixed by the code change to getimpline().  Microsoft doc says that -Oa is silently ignored.

With these changes, a VISUALC build will no longer use deprecated and confirmed erroneous optimization but will continue to use all of the other optimizations implied by -Ox.  The VISUALC build on Windows 10 host using Visual Community still fails, but gets much farther.


jmalak:
Are you saying that all standard optimization options as -O1, -O2 and -Ox are defunc and can not be used anymore? 
If yes where you get this information.
Anyway it is Microsoft problem not OW, I did fix for this MS bug in OW code.

You should use compiler which doesn't exhibit bugs with OW code, by example 19.23.28106.4 or older until Microsoft fix this bug. Only solution for MS buggy compiler is to use -Od. If you want to use 19.24.28316 then it is your decision and you can fix it localy by -Od or by whatever you want.
I don't want to do such optimization options change in OW, no reason for it.

ideafarm:
Yes, -Od also fixes VISUALC, but I kept all of the optimizations that are in -Ox that are not deprecated, so as to make the change as narrow as possible.  If you don't accept these changes, people will not be able to build open-watcom/open-watcom-v2 using Microsoft's current version of its toolset.

Feel free to direct me to assist you by making changes or by running tests.  I want you to fix this, but don't mind if you fix it in a way other than what I've done.  For example, I can put the environment variable setting in a different file.  (Which one would you think is best?)

jmalak:
Don't submit such changes, it is unacceptable.
You have damaged your build environment and trying correct it by hardcoding build system for Windows.

PR #604, opened 2020-06-02
Added a meaningful description to psp.h

davide99:
A longer description of this structure can be found [here](https://en.wikipedia.org/wiki/Program_Segment_Prefix).


jmalak:
Thanks for your contribution

PR #626, opened 2020-08-31
fix `'#pragma' is not recognized...` error when using `setvars.bat`

rivy:
Using `setvars.bat` causes spurious warning/error text when using OpenWatcom as compiler/host for builds...

```shell
C:>setvars.bat
'#pragma' is not recognized as an internal or external command,
operable program or batch file.
'#pragma' is not recognized as an internal or external command,
operable program or batch file.
Open Watcom build environment (WATCOM version=1300, CYEAR=2020)
```

This commit fixes that by only executing the "set ..." lines within 'getversi.bat' (though only on Windows NT+ machines).

### DIFF
--- a/cmnvars.bat
+++ b/cmnvars.bat
@@ -59,7 +59,8 @@ goto toolsver
 :no_intel
 :toolsver
 if not exist getversi.bat goto no_toolsver
-call getversi.bat
+if not "%OS%" == "Windows_NT" call getversi.bat
+if "%OS%" == "Windows_NT" for /f "tokens=1,*" %%G in (getversi.bat) do if /i "%%G"=="set" set %%H
 del getversi.*
 :no_toolsver
 


jmalak (commented):
It looks like you doesn't have proper setup for OW build system.
Build system suppose that bootstrap compiler is properly setup before you initiate OW build system.
Best  solution is to add as first line in setvars script command which properly setup bootstrap compiler.
What bootstrap compiler and version you use and what OS version?

jmalak:
I check it and it report this error, but problem is with C compiler preprocessor.
I is bug in preprocessor that emit pre-include special file to preprocessed output.

rivy:
It's *this* compiler.

> Open Watcom build environment (WATCOM version=1300, CYEAR=2020)

jmalak:
I created bug report #627 to fix this C compiler issue.

rivy:
Thanks.

PR #632, opened 2020-09-30
Do not parse AR environment variable when wlib is in ar mode

ArmstrongJ:
Remove a single request for the AR environment variable in wlib because it is commonly used in Makefiles as the command to call "ar" as opposed to a method for passing arguments.


ArmstrongJ:
Pull #633 supersedes this request

Issue #631, opened 2020-09-30
wlib in ar mode tries to parse AR environment variable


ArmstrongJ:
When *wlib* is used in *ar* mode, the executable will look for an environment variable *AR* and attempt to parse it as options to the library manager.  This processing occurs at bld/nwlib/c/cmdline.c:703.  It appears that this behavior is inherited from the non-*ar* mode using the *WLIB* environment variable to pass options from the IDE to the library manager (or at least that's the most common behavior).

This behavior in *ar* mode leads to problems with makefiles that will often use *AR* as an environment variable to define the command needed to execute the library manager.  The library manager will then fail with a ridiculous error such as:

```
Error! Invalid option '-/'
```

when the full path to Open Watcom's library manager is set as *AR* in the makefile.  

I feel like the *AR* environment variable should *not* be parsed by the library manager.  If others agree, I'll change it.

jmalak:
I don't know if anybody use wlib tool in ar mode and if it was working.
Maybe best solution is remove ar mode at all.
If necessary then we can create ar driver for wlib with well defined behavior.

ArmstrongJ:
I use it, but I feel like processing an environment variable in AR mode is not desired behavior.  I honestly can't think of a use case.  I'm not sure why this "feature" was ever added as it breaks so many Unix-like Makefiles.

However, providing an ar mode for *wlib* is pretty valuable if you're trying to build DOS software in a POSIX environment.  Modifying Makefiles that expect an ar-like syntax simply won't work with wlib's old Microsoft-workalike syntax, so the ar mode is quite valuable.

I am currently using the POSIX driver (*owcc*) and *wlib* in ar mode to work with a DOS port of Lua.  Other than this bug, it actually works great in that use case because I just need to redefine CC, AR, and CFLAGS to work with Open Watcom's tools rather than build a custom Makefile or fire up the IDE.  

Again, this bug breaks ar mode in a large number of cases, but *wlib* does, in fact, work well in ar mode.

ArmstrongJ:
I've created pull request #632 that basically fixes the current faulty behavior.  If *wlib*'s ar mode is so rarely used, I honestly doubt anyone was passing flags to *wlib* via the AR environment variable.  Thoughts?

jmalak:
OK. It has sense.

jmalak:
Maybe better to rename this variable to some useful name and document it instead of removing it at all.
[...]

PR #633, opened 2020-09-30
Support wlib's ar mode when the executable is named 'owar'

ArmstrongJ:
To work similar to *owcc*, the librarian will now switch to ar (POSIX) mode when the filename is *owar*.  The librarian will read options from the environment variable *OWAR* when in this mode, effectively fixing #631 while maintaining consistency with other tools.

Additionally, ar mode will be properly enabled even when the librarian is named *owar.exe*.

The name *owar* has been placed in a new header that also contains consistent environment variable definitions.

### DIFF
--- a/bld/nwlib/c/cmdline.c
+++ b/bld/nwlib/c/cmdline.c
@@ -41,10 +41,6 @@
 #include "clibint.h"
 #endif
 
-
-#define AR_MODE_ENV  "WLIB$AR"
-#define AR_MODE_ENV2 "WLIB_AR"
-
 #define eatwhite( c ) while( *(c) != '\0' && isspace( *(unsigned char *)(c) ) ) ++(c);
 #define notwhite( c ) ( (c) != '\0' && !isspace( (unsigned char)(c) ) )
 
@@ -695,12 +691,12 @@ void ProcessCmdLine( char *argv[] )
     operation   ar_mode;
 
     fname = MakeFName( _cmdname( buffer ) );
-    if( FNCMP( fname, "ar" ) == 0 || WlibGetEnv( AR_MODE_ENV ) != NULL ||
+    if( FNCMP( fname, WLIB_AR_MODE_FN ) == 0 || WlibGetEnv( AR_MODE_ENV ) != NULL ||
         WlibGetEnv( AR_MODE_ENV2 ) != NULL ) {
         Options.ar = true;
     }
     if( Options.ar ) {
-        env = WlibGetEnv( "AR" );
+        env = WlibGetEnv( "OWAR" );
     } else {
         env = WlibGetEnv( "WLIB" );
     }
--- a/bld/nwlib/c/ideentry.c
+++ b/bld/nwlib/c/ideentry.c
@@ -248,7 +248,7 @@ void Usage( void )
         MsgGet( str++, buff );
         if( Options.ar ) {
             str_last = MSG_USAGE_AR_BASE + MSG_USAGE_AR_COUNT;
-            ConsoleMessage( buff, "ar" );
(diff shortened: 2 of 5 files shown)


jmalak:
I create patch against current version of OW 2.0 which handle all what is necessary.
Try to check.

[patch-1.zip](https://github.com/open-watcom/open-watcom-v2/files/5309698/patch-1.zip)

jmalak:
I applied all necessary bug fixes for ar mode in librarian and introduced getWlibModeInfo and setWlibModeInfo
which are counter parts. setWlibModeInfo is located in IDE driver and getWlibModeInfo IDE body, both function pass data over environment.
Now it should work with old behaviour and old environment variables for reference.

Now they can be simply extended for next name "owar".

ArmstrongJ:
We can't use *AR_MODE_ENV* as an indicator on most UNIX systems.  That environment variable is defined as **WLIB$AR**, which is invalid on, for example, Linux because it contains a dollar sign.  That environment variable was the original trigger to use ar mode, but we fixed it in 2014 with issue #100 that added *AR_MODE_ENV2* defined as a more reasonable **WLIB_AR** that will work basically everywhere.  We should not be extending or using *AR_MODE_ENV* outside of backwards compatibility.

So I would say that your changes to maindrv.c should be setting *AR_MODE_ENV2* instead so that it works on platforms outside of DOS/Windows/OS2.

Second, with this patch, we now have to document that if an environment variable **WLIB_AR=1**, parse the environment variable **AR**, and if set as **WLIB_AR=2**, parse the environment variable **OWAR**.  I don't think this is optimal and adds unnecessary documentation.  It makes more sense to add another environment variable check, **WLIB_OWAR**, to force parsing the **OWAR** environment variable instead.

Quite frankly, the fact that it chooses to arbitrarily parse an environment variable named **AR**, which is exceptionally common, is absurd.  I understand the idea that it should match "behavior" of other Watcom tools, but it really was a horrible decision in this case because AR is auto-defined by GNU Autotools in almost every circumstance.  

Perhaps it would be better if this tool no longer supported an "ar" mode, but, rather, an "owar" mode.  We might want to consider dropping support entirely for *wlib* be renamed *ar* and behaving differently.  Rather, it could just respond that way when named *owar*.  I don't know how many people are using this mode, but it never should have used the generic *ar* command at all since our other tooling doesn't respond directly to being named *cc* or *ld*.  

I feel like the above change wouldn't break much.  Just a thought.

jmalak:
Sorry my mistake, I will exchange them.
But I try to return to one of my previous reply then we could build ar driver for wlib or two tools ar and wlib tools that this horrible environment variables could be removed completely and for ar we could use OWAR variable.

ArmstrongJ:
I prefer the idea of a separate *owar* tool entirely and eliminating these environment variables!

I may have misunderstood a previous reply when you suggested eliminating ar mode entirely.  I think you meant to eliminate this environment variable mess and create a new tool, *owar*, that could act like a UNIX *ar* and call *wlib* appropriately, similar to how *owcc* wraps calls to *wcc*.  Is that what you're suggesting?

jmalak:
Yes first idea was owar as ar driver for wlib or we can have two independent executable.
Wlib with this messy variables for backward compatibility is ar.exe symbolic link is defined to wlib.exe that it works and new independent tool owar as standalone tool independent on wlib (only share same source code).

jmalak:
I fixed problem with symbolic link name, now librarian uses source name instead of target name that properly identify used alias name or real name if executable is renamed.
I added owar alias that both "ar" and "owar" can be used.
We have working librarian in ar mode now, I hope.

Next we should do decision about "owar" if we create extra specialized executable or some driver to wlib mode or stay with existing solution.

jmalak:
fixed

PR #636, opened 2020-10-17
owcc option to enable emulated, inline floating-point math

ArmstrongJ:
This pull request just introduces a new option to *owcc*:

    -fpmath=emu

that will effectively pass the *-fpi* switch to *wcc*/*wcc386*.  Includes doc updates.


jmalak:
I think this option is confusing.
-fpmath= option define FPU type/optimization.

Compilers option -fpi/-fpi87/-fpc doesn't define any FPU type and -fp option for define FPU type can be combine with these options. 
By example you can combine -fp686 with -fpi.
It means optimization for 686 FPU (timming) and 387 FPU instruction set, same 387 instruction set is used by emulator.
-fpi and -fpi87 generate the same FPU code, only -fpi generate FPU instructions with fixupp to enable switch between emulator (INT call) and inline instructions.
-mhard-float should be -fpi87 (-fpi87 is default that nothing is specified now in owcc)
-msoft-float is -fpc
???? is for -fpi

I think owcc could use something like -mhard-emu-float.

ArmstrongJ:
I honestly don't care about what the flag is, as long as I can trigger passing _-fpi_ to _wcc_ since it seems to be the only way to get emulated floating point to actually work even though the docs say _wcc_ should use emulation by default.

I'll do a search/replace to change the flag to *-mhard-emu-float*, though I don't know why it would be "hard."

jmalak:
Take into account that generated code with -fpi is same as with -fpi87, try disassemble code generated with -fpi and -fpi87.
Difference between -fpi and -fpi87 is only FPU fixups, that generated code with -fpi is as -mhard-float (FPU inline instruction).

PR #637, opened 2020-10-27
Actually Implemented -mstack-size Flag in owcc

ArmstrongJ:
The documentation and command-line help both say that the flag *-mstack-size* should work for the *owcc* POSIX wrapper.  However, it doesn't appear to have ever been implemented.  This small change adds support for the flag as the documentation and command-line help both say it should.


jmalak:
Thanks for contribution.

PR #639, opened 2020-11-05
Add a "#warning" preprocessor directive

thentenaar:
This adds #warning for both cc, an plusplus.

### DIFF
--- a/bld/cc/c/cmac2.c
+++ b/bld/cc/c/cmac2.c
@@ -53,6 +53,7 @@ static void    CElse( void );
 static void    CEndif( void );
 static void    CUndef( void );
 static void    CLine( void );
+static void    CWarning( void );
 static void    CError( void );
 static void    CIdent( void );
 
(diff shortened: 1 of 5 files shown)


jmalak:
Must make changes, I have the following comments.
- #warning is not defined by any C / C ++ standard, it must be defined as a language extension, not a standard function
- the hash for #warning is too stupid (dangerous), it must be extended to all 3 values per keyword
- symbol names are a little strange, it should be rather self explanation symbols (by example user_msg symbol for function processing #error and #warning directives)

jmalak:
I were thinking about it a little bit.
I don't think that we should implement this.
OW has similar functionality by "#pragma message" which is more suitable then absolutely unportable things as non-standard pre-processor directive `#warning`. 
I don't want follow some stupidity from GCC.

fsb4000:
C23 and C++23 added `#warning`

https://en.cppreference.com/w/cpp/preprocessor/error

https://en.cppreference.com/w/c/preprocessor/error

jmalak:
Thanks for info about C23 standard which include this preprocessor directive to C/C++ standard.
In this case we can include it as extension for lower standards.

jmalak:
Little reworked for better integration as extension.
Submitted as d9181a34

PR #640, opened 2020-11-05
C Parsing Fixes

thentenaar:
This PR contains small fixes for various parsing bugs, most of which cause the compiler to spin.


jmalak:
it should be divided to more changes, it is mixture of various changes which can not be submit as one change to repository.

PR #641, opened 2020-11-05
Add support for C99 Hex float literals and binary exponents

thentenaar:


jmalak:
Be careful, you are trying to remove old compiler code (it is there for very long time).
This feature can be used by user code, that if removed it could break users code base.
Think about backward compatibility.
It is active only as extension it is not used for ISO C compatible mode.
If you are trying to remove such code then comment it out with appropriate comment about reason to be able return it back or rework code in case users will report it as problem.

PR #642, opened 2020-11-07
cc: Fix a parsing bug related to misuse of a typedef

thentenaar:
In SymLeaf() if we have a symbol that happens to be a typedef, SymLeaf()
fails to grab the next token. Thus, upon return to Statement(), the
compiler gets stuck in an infinite loop hoping for T_RIGHT_PAREN to
magically transform into T_SEMI_COLON.

### DIFF
--- a/bld/cc/c/cexpr.c
+++ b/bld/cc/c/cexpr.c
@@ -415,6 +415,7 @@ static TREEPTR SymLeaf( void )
             }
             if( sym.attribs.stg_class == SC_TYPEDEF ) {
                 CErr2p( ERR_CANT_USE_TYPEDEF_AS_VAR, Buffer );
+                NextToken();
                 return( IntLeaf( 0 ) );
             }
         }


jmalak:
The dividing of changes to small relevant parts looks OK, thanks.

One general note, OW source code has [Source Code Style Guidelines](https://github.com/open-watcom/open-watcom-v2/wiki/Source-Code-Style-Guidelines).
Please reformat all your changes to follow OW rules or you can follow style of code into which you do changes (still not all code follow these rules).
Anyway consult the guide if anything is not clear or ask here.

jmalak (changes requested):
only second place is correct
first one is for already read token then 2 tokens will be read with this change

thentenaar:
That's correct. The 2nd one is probably left over from before I refactored that function, thus I had forgotten it was there. I've removed it from the commit.

PR #645, opened 2020-11-07
cc: Fix a parsing bug in IndexOp()

thentenaar:
IndexOp() was not consuming T_RIGHT_BRACKET when given a tree rooted in
an error node, thus causing the compiler to spin.


jmalak:
fixed by [commit](https://github.com/open-watcom/open-watcom-v2/commit/f0dd4711adf150d787c401865b76153ed52bd69a)

PR #646, opened 2020-11-07
cc: Fix a bug when a block has 'else' without 'if'

thentenaar:
Statement() didn't check for T_ELSE, thus it spins looking for
T_SEMI_COLON, never reaching the end of the statement.


jmalak:
reworked and submitted as 6b497b19d94f96ec709b813d5f6e774a5ad442b6

PR #648, opened 2020-11-07
cc: Refactor duplicated code in SymLeaf()

thentenaar:


jmalak:
Code is OK.
Please reformat change to follow OW rules.

PR #649, opened 2020-11-07
cc: Scan hex float literals and update pp-number

thentenaar:


jmalak (changes requested):
In doScanFloat processing of hexadecimal format fraction part is missing.

jmalak (changes requested):
to work with new format is necessary to implement support of new format in __Strtold function in C run-time library (math library)

PR #652, opened 2020-12-07
fix including both time.h and sys/time.h for os/2.

sezero:


sezero:
There is still the problem of different struct member names: os2/sys/time.h
has ts_sec and ts_nsec, whereas other headers have tv_sec and tv_nsec

jmalak:
it is more global problem, OW standard headers contains also POSIX sys/time.h header file
question is if we should remove one of sys/time.h or merge both to one

jmalak:
Next problem is that ISO C11 and POSIX standards use following definition for timespec
 struct timespec {
     time_t    tv_sec;
     long       tv_nsec;
 };

But IBM uses incompatible definition for timespec structure with different member names
struct timespec {
   time_t  ts_sec;
   long    ts_nsec;
};                                                                                                                      

OW uses POSIX names in time.h (compiler version).

it has no simple solution, possibility are (only for OS/2)
- use POSIX naming and not be compatible with IBM
- use IBM naming and not be compatible with POSIX
- use POSIX naming and use macros for IBM naming (map ts_... -> tv_... )

sezero:
The following using a union seems to work and covers both os/2-specific and posix cases:
```
struct timespec {
  union {
    struct {
      time_t  ts_sec;
      long    ts_nsec;
    };
    struct {
      time_t  tv_sec;
      long    tv_nsec;
    };
  };
};
```

sezero:
> The following using a union seems to work and covers both os/2-specific and posix cases:
> 
> ```
> struct timespec {
>   union {
>     struct {
>       time_t  ts_sec;
>       long    ts_nsec;
>     };
>     struct {
>       time_t  tv_sec;
>       long    tv_nsec;
>     };
>   };
> };
> ```

Do you agree with a solution like this?

jmalak:
perfect, anonymous structures are much better solution then macros
I am not sure a problem with strict ansi mode (-za option) could be there (anonymous structures/unions are not part of any standard)

sezero:
> perfect, anonymous structures are much better solution then macros
> I am not sure a problem with strict ansi mode (-za option) could be there
> anonymous structures/unions are not part of any standard, probably implemented as extension

Yes, it does fail with `Error! E1026: Invalid declarator` errors if -za is specified.
We can conditionalize anon struct/union usage with `NO_EXT_KEYS` ifdefs
(making it more ugly..)

Do you want a patch (or a new PR), or will you handle things?

jmalak:
If you agree I will prepare change request for comment
It will include all necessary changes in appropriate header files to work for OS/2 with both member names
All other platforms will be same as now.
The problem is that IBM development pack is very old and was created with sys/time.h from old BSD version
Any solution is acceptable even if it doesn't look nice

sezero:
> If you agree I will prepare change request for comment

OK, I will review/test your patches

sezero:
Closing in favor of #656

Issue #752, opened 2021-11-06
Open Watcom binaries cannot stat() anything on 64-bit Linux (x86_64) on a filesystem with device node numbers higher than 256


joncampbell123:
There seems to be a problem on a newer laptop of mine that is a combination of several factors as far as I can tell:
- The kernel is 64-bit
- Open Watcom compiles itself as i386 32-bit
- Open Watcom tools use stat() to check for the existence of files or examine their status
- The filesystem was mounted from an NVME (Non-Volatile Memory Express) storage device, which has device node numbers higher than 255: (259, 6) in this case.
- stat() for any file fills those device node numbers into st_dev
- The Linux i386 compatibility layer for 64-bit kernels will fail stat() with EOVERFLOW if any member of stat() on return is too large for the structure returned to userspace, including the inode, dev, and rdev fields, and of course the file size. There is code in sys_newstat(), the function behind SYS_stat, to return that error if the device node is "not valid" i.e. either number is 256 or higher.

On my other laptop, Open Watcom works just fine because I am running it off of an SSD connected through a SATA controller, device node numbers (8, 6) which is well below 256.

One symptom of this issue is that wmake acts as if the source file to compile into an object doesn't exist, then complains that it cannot make the target.

In DOSLIB, for bootsect/i13geom/boot1.asm, wmake stat()s the file, but even though the ASM file is well below 2GB (about 4KB in fact!), the stat() call returns EOVERFLOW.

joncampbell123:
This is something to take seriously because it means anyone running Linux on a newer laptop with NVME storage (as most laptops are now) will be unable to do anything with Open Watcom at all.

joncampbell123:
Here is where stat() leads to in the Linux kernel, or rather, what it uses to copy the structure to userspace and validate it before returning to userspace.

https://elixir.bootlin.com/linux/latest/source/fs/stat.c#L337

https://elixir.bootlin.com/linux/latest/source/fs/stat.c#L344

Note the !valid_dev() checks. When compiled again to run in the 32-bit compatibility API, valid_dev leads to old_valid_dev() which leads to a check for whether device numbers are below 256:

https://elixir.bootlin.com/linux/latest/source/include/linux/kdev_t.h#L24

NVME devices end up with major 259, which fails that check.

joncampbell123:
Note that this doesn't affect anything linked against GLIBC because GLIBC will automatically use newer stat functions not limited in this way, whether 32-bit or 64-bit.

[...]

PR #753, opened 2021-11-07
Linux stat64 workaround

joncampbell123:
This change modifies stat() on Linux so that stat() uses the stat64() syscall to work around limitations with the original stat() call and modern 64-bit kernels.

https://github.com/open-watcom/open-watcom-v2/issues/752


joncampbell123:
Yikes! Hold on a second...

PR #754, opened 2021-11-07
On Linux, call stat64() even from the stat() function to work around limitations with the i386 Linux kernel stat syscall on 64-bit kernels.

joncampbell123:
This is a fix for https://github.com/open-watcom/open-watcom-v2/issues/752

### DIFF
--- a/bld/clib/linux/c/stat.c
+++ b/bld/clib/linux/c/stat.c
@@ -34,18 +34,8 @@
 #include <sys/stat.h>
 #include "linuxsys.h"
 
-#define DEFER_TO_STAT64
-
-#ifdef DEFER_TO_STAT64
-# define _BSD_SOURCE
-# include <sys/types.h>
-
-# define linuxkernel_old_dev_valid(x) (major(x) <= 0xFF && minor(x) <= 0xFF)
-#endif
-
 _WCRTLINK int stat( const char *filename, struct stat * __buf )
 {
-#ifdef DEFER_TO_STAT64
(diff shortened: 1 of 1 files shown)


jmalak:
Please remove guard for conditional code, I think it is useless.
Remove _BSD_SOURCE macro too.

joncampbell123:
_BSD_SOURCE is needed for the major() and minor() macros when checking st_dev and st_rdev.

It is wrapped in an #ifdef so you can comment out the #define and get the old stat() behavior back very easily.

Should I still remove the guard, make the stat64 behavior the only behavior?

jmalak:
I think it is not covered by POSIX.
I don't want relate to BSD any way for Linux implementation.
What about <sys/sysmacros.h> on Linux to map dev ID's?

jmalak:
To the guard, I don't think that we will return back to old functionality.
There is detailed description why it is done that if it is real problem then we can remove it  or change it somehow.
The use of stat system call is included if stat64 fails that it is OK.

joncampbell123:
sys/sysmacros.h is in my /usr/include/linux directory but Open Watcom doesn't see it while compiling itself, so I'll remove the major, minor checks. I think both stat and stat64 use the same data type for st_dev and st_rdev so it should be OK to copy through. The kernel stat() limits to 255, 255 for legacy reasons, but we don't have to.

I'll remove the conditional part.

jmalak:
It is not a problem. I will remove appropriate macros from sys/types.h and add sys/sysmacros.h to OW to use same code base and users could use it to correct manipulation with major/minor number of device.

PR #800, opened 2022-01-16
Fix a typo for option mapping:  -fomit-fp-rounding

winspool:
Signed-off-by: Detlef Riekenberg <wine.dev@web.de>


jmalak:
Thanks for your contribution

PR #817, opened 2022-02-18
Fixed vector insert taking an iterator defined range. Fixes #197

### LINKED ISSUE #197: Bugs in vector::insert()
Function `vector::insert(iterator pos, Iter first, Iter last)` has a bug. Function `helper_insert` use this loop for inserting each item [first, last):
```c++
        while( first != last ) {
            insert( position, *first );
            ++position;
            ++first;
        }
```
Function call `insert(position, *first)` can reallocate vector buffer, and then iterator `position` will be avoid. For fixing use `position = insert(position, *first);`.

Other bug in function `vector::insert(iterator position, const Type& x)` — case of `position == end()` is not implemented.

pchapin:
This PR includes new tests that demonstrate the issue. With the fix, those new tests are now working. I was unable to reproduce the second bug mentioned in the issue. A new issue can be created for that bug if necessary.

### DIFF
--- a/bld/hdr/readme.txt
+++ b/bld/hdr/readme.txt
@@ -44,12 +44,11 @@ maintenence.
 
 Headers for most of the target systems (DOS, Windows, WinNT, OS/2) are
 generated into the 'dos' directory. Headers for Linux are generated into
-the 'linux' directory. Header for RDOS are generated into the 'rdos' directory.
-Headers for QNX are disabled by default. When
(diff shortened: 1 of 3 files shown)


pchapin:
Hi Jiri, I think you forgot to merge this? I assume you intended to merge it, but it was closed without being merged. I can do it myself one the checks are complete (again).

jmalak:
Oh sorry, I am stupid.

PR #854, opened 2022-04-24
[wcl/owcc] Quote filenames with @ symbols

Ronsor:
This patch ensures that filenames and strings in general that contain the `@` symbol are quoted, allowing files with that symbol in the name to be compiled successfully. Without it, WLINK misinterprets the names.

### DIFF
--- a/bld/wcl/c/clcommon.c
+++ b/bld/wcl/c/clcommon.c
@@ -445,7 +445,7 @@ char *DoQuoted( char *buffer, const char *name, char quote_char )
 
     p = buffer;
     if( name != NULL ) {
-        quotes = ( strchr( name, ' ' ) != NULL );
+        quotes = ( strpbrk( name, " @" ) != NULL );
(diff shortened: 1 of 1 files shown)


jmalak:
It is more general issue with host system valid file name character set and wlink command line parser.
OW support mostly only DOS file name valid character set that issue can be with other characters on different hosts.
Probably wlink parser should be fixed to handle file names more portable way and don't restrict file name to DOS file name character set.
Please could you give me example of wcl/owcc command line for this issue and on what host system you used it.

jmalak:
I submit change 2f189c43b43177171d9677b292e60a2d54aa47fd for wlink which should solve this issue.
Now wlink include '@' character as valid character for any path/file name on any host system.
It resolves this issue also for wlink and wfc.. utilities.

Ronsor:
I can confirm it works perfectly now. Thanks.

PR #916, opened 2022-07-02
Fix memory errors in wasm

DelilahHoare:
This will be needed on osx arm.

### DIFF
--- a/bld/wasm/c/macro.c
+++ b/bld/wasm/c/macro.c
@@ -103,7 +103,7 @@ static char *replace_parm( parm_list *parms, char *start, size_t len, asmlines *
 
             new_line = AsmAlloc( strlen( old_line ) - len + PLACEHOLDER_SIZE + 1 );
             before = start - old_line;
-            if( *(start - 1) == '&' )
+            if( ( start > lstruct->line ) && ( *(start - 1) == '&' ) )
                 before--;
             strncpy( new_line, old_line, before );
             *(new_line + before) = '\0';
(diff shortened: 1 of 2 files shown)


jmalak (commented):
comparison includes null termination byte, therefore this situation cannot happen

DelilahHoare:
> comparison includes null termination byte, therefore this situation cannot happen

memcmp is allowed to read anywhere in the passed ranges, and the popular implementations do use SIMD instructions which can read far past the terminator.  The call causes an error when running with ASAN.

jmalak:
> > comparison includes null termination byte, therefore this situation cannot happen
> 
> memcmp is allowed to read anywhere in the passed ranges, and the popular implementations do use SIMD instructions which can read far past the terminator. The call causes an error when running with ASAN.

It can not compare data after min( len of string1, len of string2 ) + 1 because it always end on first "null" in any of both strings.

jmalak:
> > > comparison includes null termination byte, therefore this situation cannot happen
> > 
> > 
> > memcmp is allowed to read anywhere in the passed ranges, and the popular implementations do use SIMD instructions which can read far past the terminator. The call causes an error when running with ASAN.
> 
> It can not compare data after min( len of string1, len of string2 ) + 1 because it always end on first "null" in any of both strings. How implementation read and process data is not important. C standard is defining behavior of memcmp exactly and implementation must respect it otherwise it is nonconforming implementation.

I looked into standard and you are right, standard doesn't require to stop on first difference. It allows to read full length on both arguments.

jmalak (commented):
I think following code could be more transparent
```
            before = start - old_line;
            if( before > 0 ) {
                if( *(start - 1) == '&' )
                    before--;
                strncpy( new_line, old_line, before );
            }
            *(new_line + before) = '\0';
            if( sprintf( buffer, "#%-2.2d", count ) != 3 ) {
                myassert( 0 );
            }
            strcat( new_line, buffer );
```

PR #928, opened 2022-07-21
Add support for FM TOWNS's version of Phar Lap

trap15:
The Fujitsu FM TOWNS computer series uses Phar Lap for nearly all software built for it. However, this seems to be an older version than what is currently supported, and so the version check in the startup code was failing to properly handle it.

Additionally, the INT 21h,AH=63h DOS call behaves differently than was expected, causing crashes in initialization.

I haven't done particularly extensive testing of the rest of clib or the functions that use the extender, but I'm able to at least print a hello world now. If I encounter more issues I'll make follow up PRs.

### DIFF
--- a/bld/clib/mbyte/c/getltdos.c
+++ b/bld/clib/mbyte/c/getltdos.c
@@ -117,7 +117,16 @@ unsigned short __far *dos_get_dbcs_lead_table( void )
         segregs.ds = _FP_SEG( &pblock );
         intdosx( &regs, &regs, &segregs );
         if( regs.x.cflag == 0 && pblock.real_eax.b.l == 0 ) {
-            if( pblock.real_ds != 0xFFFF ) {    /* weird OS/2 value */
+            if( pblock.real_ds == 0xFFFF ) {    /* weird OS/2 value */
+                return( NULL );
+            } else if( pblock.real_ds == 0 && regs.w.si == 0) {
+                /* Because EXTENDER_RM2PM forces the segment value, without this
+                 * a NULL check on the result of this function would never detect it
+                 * so long as this DOS call doesn't fail.
+                 * At least FM TOWNS returns success but DS:SI=0:0 (which is bogus memory).
+                 */
+                return( NULL );
+            } else {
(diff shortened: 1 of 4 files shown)


bld/clib/mbyte/c/getltdos.c
```diff
@@ -117,7 +117,16 @@ unsigned short __far *dos_get_dbcs_lead_table( void )
...
         intdosx( &regs, &regs, &segregs );
         if( regs.x.cflag == 0 && pblock.real_eax.b.l == 0 ) {
-            if( pblock.real_ds != 0xFFFF ) {    /* weird OS/2 value */
+            if( pblock.real_ds == 0xFFFF ) {    /* weird OS/2 value */
+                return( NULL );
+            } else if( pblock.real_ds == 0 && regs.w.si == 0) {
+                /* Because EXTENDER_RM2PM forces the segment value, without this
+                 * a NULL check on the result of this function would never detect it
+                 * so long as this DOS call doesn't fail.
+                 * At least FM TOWNS returns success but DS:SI=0:0 (which is bogus memory).
```
trap15, line 126: The return from this call is all registers cleared, except EAX is left at 6300h. I believe this is probably the most appropriate solution to this result.

bld/clib/startup/a/cstrt386.asm
```diff
@@ -202,9 +202,17 @@ ENV_SEG equ     2ch
         shr     eax,16                  ; get top 16 bits of eax
         cmp     ax,'DX'                 ; if top 16 bits = "DX"
         jne     not_pharlap             ; then its pharlap
+        ;cmp    ebx,'12aJ'              ; - if ebx (version number) is 12aJ
+        cmp     ebx,4A613231h           ; - if ebx (version number) is 12aJ
```
trap15, line 206: This appears to be the only version number ever used by the FM TOWNS version of Phar Lap, it at least works with the built-in version on the last version of the OS.

jmalak:
Thanks for contribution.
It was reimplemented to the current version of OW

PR #949, opened 2022-09-07
cmac1: Fix buffer overflow during macro expansion

jackrosenthal:
As I figured out while compiling MicroPython with Open Watcom, the comment here "size of Buffer is OK" is a lie, as I noticed wcc segfault.

Unfortunately, I wasn't able to derive a minimal example which replicates it, but here are steps you can use to reproduce the bug using the same source I was (this assumes you have WATCOM and related environment variables set correctly):

    $ git clone https://github.com/micropython/micropython
    $ cd micropython
    $ git checkout 4e4c28bf27ff4a9e1a2ce5b6fffa641d9a332507
    $ mkdir ports/dos
    $ cd ports/dos
    $ wcc -bt==dos -I. -I../.. -Ibuild \
          -DMICROPY_ROM_TEXT_COMPRESSION=1 -DNO_QSTR -pl -0 -zq \
          -fr ../../py/objdict.c
    ~> Segmentation fault

Using GDB and some hardware watchpoints, I was able to deduce the segfault was caused by the scanbuf (a.k.a., the "Buffer" global variable) overflowing and corrupting the following MCB, and it was being happening at the code I modified in this commit.

This changes that code to instead using WriteBufferChar and WriteBufferNullChar, which will bounds check Buffer and re-alloc when required.

### DIFF
--- a/bld/cc/c/cmac1.c
+++ b/bld/cc/c/cmac1.c
@@ -248,6 +248,7 @@ TOKEN GetMacroToken( void )
     MACRO_TOKEN     *mtok;
     bool            keep_token;
     TOKEN           token;
+    char            c;
 
     for( ; (mtok = TokenList) != NULL; ) {
         if( (token = mtok->token) != T_NULL ) {
@@ -265,10 +266,11 @@ TOKEN GetMacroToken( void )
         token = T_NULL;
         Buffer[0] = '\0';
     } else {
-        /* size of Buffer is OK, token data was processed in Buffer before */
-        while( (Buffer[TokenLen] = mtok->data[TokenLen]) != '\0' ) {
-            TokenLen++;
+        while( (c = mtok->data[TokenLen]) != '\0' ) {
+            WriteBufferChar(c);
         }
+        WriteBufferNullChar();
+
         keep_token = false;
         switch( token ) {
         case T_SAVED_ID:


jmalak (commented):
Many thanks for your contribution.
Please give me a time to review OW token processing.
It could affect another token processing if the comment in previous code is not valid.

jmalak:
> Many thanks for your contribution. Please give me a time to review OW token processing. It could affect another token processing if the comment in previous code is not valid.

I am sorry I cannot reproduce your issue.
Anyway the output is different from me, because you use preprocesing **`-pl`** option.
Please could you give me correct command line for wcc.

jmalak:
> wcc -bt==dos -I. -I../.. -Ibuild \
>       -DMICROPY_ROM_TEXT_COMPRESSION=1 -DNO_QSTR -pl -0 -zq \
>       -fr ../../py/objdict.c

I tried to compile it, not preprocess, but there is lot of errors.
I need your mpconfigport.h file.
Anyway code looks like C99 that you need add -za99 option for compiler.
Compiler should not crash even if you use incorrect options.

jackrosenthal:
Thanks, I should have tested the steps I put in the commit message!

I updated the commit message with new steps:

        $ git clone -b dos https://github.com/jackrosenthal/micropython
        $ cd micropython/ports/dos
        $ wcc -bt==dos -za99 -I. -I../.. -Ibuild \
              -DMICROPY_ROM_TEXT_COMPRESSION=1 -DNO_QSTR -pl -0 -zq \
              -fr ../../py/objdict.c
        ~> Segmentation fault

jmalak:
> wcc -bt==dos -za99 -I. -I../.. -Ibuild \
>           -DMICROPY_ROM_TEXT_COMPRESSION=1 -DNO_QSTR -pl -0 -zq \
>           -fr ../../py/objdict.c

I am not able to reproduce your issue on your repository clone.
I get regular pre-processed output not segment fault.
I tested it by latest OW CI build from today morning.
I am using Ubuntu 18.04 (64-bit version) and test it using 64-bit version of OW C compiler.
```
Open Watcom C x86 16-bit Optimizing Compiler
Version 2.0 beta Sep  7 2022 04:12:55 (64-bit)
```

I tested it also by 32-bit OW C compiler and no problem.
```
Open Watcom C x86 16-bit Optimizing Compiler
Version 2.0 beta Sep  7 2022 04:12:49 (32-bit)
```

jackrosenthal:
Did you check the return code after execution? The segmentation fault happens after the preprocessor output has finished, during memory cleanup.

jmalak:
> Did you check the return code after execution? The segmentation fault happens after the preprocessor output has finished, during memory cleanup.

You are reporting segmentation fault, but it doesn't happen.

jmalak:
> > Did you check the return code after execution? The segmentation fault happens after the preprocessor output has finished, during memory cleanup.
> 
> You are reporting segmentation fault, but it doesn't happen. Only issue could be that 64-bit version uses **`-march=core2 -msse2`** GCC compile options, but I think all 64-bit Intel/AMD CPU should be compatible with this setup.

I am able to reproduce it now, there is difference between CI and release build compiler options.

jmalak:
> > > Did you check the return code after execution? The segmentation fault happens after the preprocessor output has finished, during memory cleanup.
> > 
> > 
> > You are reporting segmentation fault, but it doesn't happen. Only issue could be that 64-bit version uses **`-march=core2 -msse2`** GCC compile options, but I think all 64-bit Intel/AMD CPU should be compatible with this setup.
> 
> I am able to reproduce it now, there is difference between CI and release build compiler options.

I checked and the Buffer processing is OK, there is not issue.
It looks like some pointer mistake to overwrite some memory block by some incorrect pointer or mistake in memory managing procedures of C compiler under some conditions. It causes segmentation fault during memory cleanup, because some MCB's are damaged. I can depend on any memory operation and depend on if some MCB is damage.
It need more investigation.

jackrosenthal:
What I did to track this down as the offending code that overwrote the MCB:

Run in GDB, and after the segfault occurs, find the address of the MCB that was overwritten. Set a hardware watchpoint on this address, and re-run. This code was where the watchpoint triggered.

Likewise, with this patch applied, I don't see the same segfault (do you still see it?)

If you're seeing something different, maybe try the watchpoint trick above to track the code that overwrote the MCB.

jmalak:
> What I did to track this down as the offending code that overwrote the MCB:
> 
> Run in GDB, and after the segfault occurs, find the address of the MCB that was overwritten. Set a hardware watchpoint on this address, and re-run. This code was where the watchpoint triggered.
> 
> Likewise, with this patch applied, I don't see the same segfault (do you still see it?)
> 
> If you're seeing something different, maybe try the watchpoint trick above to track the code that overwrote the MCB.

But the source of problem is not length of buffer it has always enough length, that code note is OK.
Source of problem is damaged MCB by overwriting by now unknown part of code.
Your fix mask only a problem by reallocating Buffer (this is useless), that you use different MCB and different memory blocks map and luckily don't damage any MCB, but it is not fix for root cause which is probably some memory damage by memory block overwriting.

jackrosenthal:
> Source of problem is damaged MCB by overwriting by now unknown part of code.

The code that overwrote the MCB was this loop. I suggest using a watchpoint in gdb on the address of the ovewritten MCB to observe this: this is how I did.

jmalak:
> > Source of problem is damaged MCB by overwriting by now unknown part of code.
> 
> The code that overwrote the MCB was this loop. I suggest using a watchpoint in gdb on the address of the ovewritten MCB to observe this: this is how I did.

It's too late, the memory is already corrupted/overwritten.

jmalak:
It is fixed by commit f40c35ec8c68025cda1e433be9d1356d534d6579

PR #1015, opened 2022-12-19
fix vxd writer to account for DDB not in the start of a segment

javispedro:
Wlink's VxD generator sets some fields of the LE header based on data from the VxD's device descriptor block (DDB). However, the current code assumes that the the DDB is always at offset 0 in some segment. 
If the DDB is not at offset 0 then it just reads random data from the DDB's segment thereby writing a corrupt VxD. 
This patch fixes this. 

(having a DDB at offset 0 is not necessary for the VxD to be loaded by windows, in fact most builtin vxds have the DDB at the end of the segment).


jmalak:
Thanks for your contribution.

PR #1022, opened 2023-01-05
coptions: Add unix compiler compatible defines for target arch and OS

winspool:
OW uses only DOS/Windows compatible defines similar to the MS compiler for the target architecture and OS and fails to compile a lot of Non-Windows software.

Add defines used by gcc to fix that.

--
Regards ... Detlef

Signed-off-by: Detlef Riekenberg  <wine\x2edev\x40web\x2ede>

### DIFF
--- a/bld/cc/c/coptions.c
+++ b/bld/cc/c/coptions.c
@@ -218,6 +218,7 @@ static void SetTargetSystem( void )
         PreDefine_Macro( "M_I86" );
 #elif _CPU == 386
         PreDefine_Macro( "M_I386" );
+        PreDefine_Macro( "i386" );
 #elif _CPU == _AXP
         PreDefine_Macro( "M_ALPHA" );
 #elif _CPU == _PPC
@@ -234,26 +235,35 @@ static void SetTargetSystem( void )
     PreDefine_Macro( "__X86__" );
     PreDefine_Macro( "_X86_" );
 #elif _CPU == 386
+    PreDefine_Macro( "__i386__" );
+    PreDefine_Macro( "__i386" );
     PreDefine_Macro( "_M_I386" );
     PreDefine_Macro( "__386__" );
     PreDefine_Macro( "__X86__" );
     PreDefine_Macro( "_X86_" );
     PreDefine_Macro( "_STDCALL_SUPPORTED" );
 #elif _CPU == _AXP
+    PreDefine_Macro( "__alpha__" );
     PreDefine_Macro( "_M_ALPHA" );
     PreDefine_Macro( "__ALPHA__" );
     PreDefine_Macro( "_ALPHA_" );
     PreDefine_Macro( "__AXP__" );
     PreDefine_Macro( "_STDCALL_SUPPORTED" );
 #elif _CPU == _PPC
+    PreDefine_Macro( "__powerpc__" );
+    PreDefine_Macro( "__ppc__" );
     PreDefine_Macro( "_M_PPC" );
     PreDefine_Macro( "__POWERPC__" );
     PreDefine_Macro( "__PPC__" );
     PreDefine_Macro( "_PPC_" );
 #elif _CPU == _MIPS
+    PreDefine_Macro( "__mips__" );
+    PreDefine_Macro( "__mips=32" );
     PreDefine_Macro( "_M_MRX000" );
     PreDefine_Macro( "__MIPS__" );
 #elif _CPU == _SPARC
+    PreDefine_Macro( "__sparc__" );
+    PreDefine_Macro( "__sparc" );
     PreDefine_Macro( "_M_SPARC" );
     PreDefine_Macro( "__SPARC__" );
     PreDefine_Macro( "_SPARC_" );
@@ -330,6 +340,7 @@ static void SetTargetSystem( void )
             PreDefine_Macro( "MSDOS" );
         }
         PreDefine_Macro( "_DOS" );
+        PreDefine_Macro( "__MSDOS__" );
         break;
 
 #if _CPU == 386
@@ -350,13 +361,31 @@ static void SetTargetSystem( void )
         break;
 
     case TS_NT:
+        if( CompFlags.non_iso_compliant_names_enabled ) {
+            PreDefine_Macro( "WIN32" );
+            PreDefine_Macro( "WINNT" );
+        }
         PreDefine_Macro( "_WIN32" );
+        PreDefine_Macro( "__WIN32" );
+        PreDefine_Macro( "__WIN32__" );
+        PreDefine_Macro( "__WINNT" );
+        PreDefine_Macro( "__WINNT__" );
         break;
 #endif
-    case TS_QNX:
-        /* fall through */
     case TS_LINUX:
+        if( CompFlags.non_iso_compliant_names_enabled ) {
+            PreDefine_Macro( "linux" );
+        }
+        PreDefine_Macro( "__linux__" );
+        PreDefine_Macro( "__linux" );
+        /* fall through */
+    case TS_QNX:
+        if( CompFlags.non_iso_compliant_names_enabled ) {
+            PreDefine_Macro( "unix" );
+        }
         PreDefine_Macro( "__UNIX__" );
+        PreDefine_Macro( "__unix__" );
+        PreDefine_Macro( "__unix" );
         break;
 
     case TS_CHEAP_WINDOWS:
@@ -745,14 +774,20 @@ static void MacroDefs( void )
     case CPU_486:
         Define_Macro( "__SW_4" );
         PreDefine_Macro( "_M_IX86=400" );
+        PreDefine_Macro( "__i486__" );
+        PreDefine_Macro( "__i486" );
         break;
     case CPU_586:
         Define_Macro( "__SW_5" );
         PreDefine_Macro( "_M_IX86=500" );
+        PreDefine_Macro( "__i586__" );
+        PreDefine_Macro( "__i586" );
         break;
     case CPU_686:
         Define_Macro( "__SW_6" );
         PreDefine_Macro( "_M_IX86=600" );
+        PreDefine_Macro( "__i686__" );
+        PreDefine_Macro( "__i686" );
         break;
     }
     switch( SwData.fpt ) {


jmalak:
Please specify what is "unix compiler compatible defines".
I don't know about any standard which specify such items.
For ISO compliant code is not necessary to use any such macros, all are related to exact compiler C language extensions etc.
Generally code which use some extension is unportable and you can not relate on such macros because they are unportable.
We have several times various discussion about request to implement GCC things (extensions etc.) not only predefined macros. We don't want to implement such non-standard things any way until we have very very good reason.
OW is not clone of any compiler as MSVC or GCC and will not be in the future.
OW supports many targets as native cross-compiler that use different set of predefined macros as another compilers and there is not reason to extend number of macros for the same things because it is never-ending story with adding things to do unportable code "more portable".

pchapin:
@winspool, you mention two different issues in your post. The first is focused on the environment macros defined (or not) by Open Watcom. The second issue is broader and is mostly about how Open Watcom lacks features from the newer standards and lacks support for various modern platforms.

In answer to the second point, I would love to see Open Watcom have broader standards and platform support, but unfortunately the small size of the development team makes that difficult.

Regarding the first point, it is good to keep in mind that the commercial Watcom compiler was never intended to support Unix-like systems at all. The support that exists, such as it is, has all been added by that same-said small development team. Although there is work being done to enhance the support for modern platforms, a large part of the Open Watcom community is focused on legacy software for which there is, essentially, no other compiler available. My sense is that group does not have a burning need for compatibility with gcc, clang, etc.

That said, I don't see a problem with adding some additional environment macros if doing so helps to increase compatibility. If they are in the implementation's name space there shouldn't be an issue for existing, legacy software.

You said in your post that `printf` is "unportable" between 16-, 32-, and 64-bit platforms (or did you mean data types?). Could you elaborate on that? I don't understand what you mean.

jmalak:
any code that uses the extension is fully compilable only by the compiler whose extension you are using and not by some other
with other compilers partial compatibility can be achieved but never full, therefore OW does not go this way
to compile code with the GCC extension, you must use the GCC compiler
to compile code with the MS Visual C extension, you must use the MS Visual C compiler
to compile code with the Watcom extension you must use the Watcom compiler
etc.
that's why I was talking about the never ending story if we try for compatibility with different expansions it's a road to hell
we're trying to improve compatibility with standards so we'll allow compiling code that doesn't use extensions
**OW is not and never will be the most widely used type of compiler**, nor will we go down the path of implementing other compiler extensions unless there is a very good reason to do so
adding these predefined macros is simple but has no meaning as it will not create the extensions for which these macros are used by the original compiler

winspool:
@pchapin, thanks for joining this pull request

> @winspool, you mention two different issues in your post. 

>The first is focused on the environment macros defined (or not) by Open Watcom. 
> The second issue is broader and is mostly about how Open Watcom lacks features 
> from the newer standards and lacks support for various modern platforms.

Oh sorry, that was not the intention for this PR 

> In answer to the second point, I would love to see Open Watcom have broader standards 
> and platform support, but unfortunately the small size of the development team 
> makes that difficult.

Yes, but you make it very difficult to join OpenWatcom development,
and without changing that, the small development team will never enlarge.

> Regarding the first point, it is good to keep in mind that the commercial Watcom compiler
> was never intended to support Unix-like systems at all. 
> The support that exists, such as it is, 
> has all been added by that same-said small development team.
> Although there is work being done to enhance the support for modern platforms,
> a large part of the Open Watcom community is focused on legacy software 
> for which there is, essentially, no other compiler available. 

I know that, and that's why i'm still here.
I even have a Win NT 3.1 (Server and some clients) license 
and a NT4.0 client license (includes x86, aplpha, mips) 
and i used them in my early wine days
 
> My sense is that group does not have a burning need for compatibility with 
> gcc, clang, etc.

But simple using CC=owcc instead of CC=tcc or CC=gcc helps me and others a lot.

> That said, I don't see a problem with adding some additional environment macros 
> if doing so helps to increase compatibility.
> If they are in the implementation's name space there shouldn't be an issue 
> for existing, legacy software.

Then go and apply this Pull Request
 
> You said in your post that `printf` is "unportable" between 16-, 32-, and 64-bit platforms > (or did you mean data types?). 

The format string is unportable for some data types.
Example is the format string for printing 64 bit data, when the target is 32bit or 64bit,
DOS/Windows or unix (gcc, clang, tcc, ..)
(i know inttypes.h, but detecting C99 compatibility fails for owcc
by autoconf: VLA not supported, see attached log)


[config_owcc_c99.log](https://github.com/open-watcom/open-watcom-v2/files/10368213/config_owcc_c99.log)

winspool:
Hi @jmalak 

> * any code that uses the extension is fully compilable only by the compiler 
>   whose extension it is using and not by some other compiler
> 
> * other compilers partial compatibility can be achieved but never full,
>   therefore OW does not go this way
> 
>  * to compile code with the GCC extension, you must use the GCC compiler
> 
>  * to compile code with the MS Visual C extension, you must use the MS Visual C compiler
> 
>  * to compile code with the Watcom extension you must use the Watcom compile
>  etc.

I know that.

> that's why I was talking about the never ending story 
> if we will try for compatibility with different extensions it's a road to hell
> **we're trying to improve compatibility with standards so we'll allow compiling code that doesn't use extensions**
> **OW is not and never will be the most widely used type of compiler**, 
> nor will we go down the path of implementing other compiler extensions
> unless there is a very good reason to do so

I know that

> adding these predefined macros as you proposed is simple 

Then please apply this Pull request

> but has no meaning as it will not create the extensions 
> for which these macros are used by the code and original compiler

It has a meaning, but you don't see it.
It has nothing to do with supporting compiler specific extensions.
It's simple about selecting a different implementation,
depending on OS (Posix, Windows, DOS) or arch.

OpenWatcom uses such defines all over its own source code for the same reason,
but the defines used by OpenWatcom are used in no other compiler.

Adding defines similar to the defines used by other compiler (tcc or gcc as examples)
helps to use OpenWatcom as an easy drop-in compiler without porting
the software to OpenWatcom.
That will work without supporting compiler specific extensions.

> Anyway standard for these predefined macros doesn't exist, each compiler use its own set.

There is no standard, but understanding the predefined marcros used by leading compiler
helps to compile other software without the requirement to port that software.
Watcom did the same in the early days (DOS), 
but OpenWatcom did not move forward. 

> Any use of such macros in code signal use of extension 
> and therefore introduce issues which I comment before.

See above. You still failed to see some reasons for such predefined marcros.

> I understand that people like "super-compiler" which can compile any code, 
> but it is unrealistic expectation.

I know that.

> OW is not GCC or CLANG which compete to get dominance it is not our objective.

I know that.

> OW can be used where "predators compilers" is not usable or with problems 
> and it is our main objective.

I know that.

Yes, the OW development team is very small,
but when someone else tries to help to improve OpenWatcom,
a warm welcome would be a much better feeling.

For example, something like this:
"Done. Thanks for the Patch, but keep in mind, that we have decided
to stay to the C/C++ standard and avoid to add compiler specific extensions"

That would be simple and fast, instead of such a long conversation.

jmalak:
Sorry but I don't agree with you, it is not simple as you present it.
In any case using such macros signal non portable code (because it is implementation specific) and there will be more problems not only missing macros of another C/C++ implementation. 
Even if we add such compiler predefined macros in OW it doesn't resolve portability issue which are in such code.
What problem is to put such macros on compiler command line?

pchapin:
I think I agree with @jmalak here. If we add the macros defined by other compilers, it makes it appear that we are trying to support the extensions of those other compilers, which we aren't. So it is misleading to developers. If feels like a hacky fix that might work in some simple cases, but not enough to be provide the sort of portability you are envisioning. A program that depends on gcc extensions, and uses gcc's macros to select for them, is unlikely to compile with Open Watcom even if Open Watcom defined the same macros. A an actual porting effort is going to be needed in any case, and in that case, the code could be edited to use the Open Watcom macros. At least, I think so?"

On the other hand, I can understand the point you're making. We do have `owcc` which provides a gcc-like interface to Open Watcom to facilitate using Unix Makefiles out-of-the-box. It could be said that providing gcc-like macros is just another example of that sort of thing. Even if it doesn't always work, it might work well enough, enough of the time, to make it worthwhile.

Yet, it could be argued that having `owcc` is a mistake for the reasons I mentioned earlier. It misleads developers into thinking we are more compatible with gcc than we are.

How about this compromise position: We change `owcc` so that when it launches the compilers it provides them with `-d` options to activate certain gcc-like macros. The compiler remains true to itself, and the compatibility tool, which is known to be imperfect, even if sometimes useful, deals with the matter.

pchapin:
@winspool I am still not entirely clear when you say:

> The format string is unportable for some data types.
> Example is the format string for printing 64 bit data, when the target is 32bit or 64bit,
> DOS/Windows or unix (gcc, clang, tcc, ..)

The best I can make out is that you're talking about a case like this:

    typedef counter_t int; 
    // ...
    counter_t counter = 0;
    // ...
    printf( "The count = %d\n", counter );

If the `counter_t` alias is later changed to `long` then the `%d` format specifier is incorrect. It will still work if `int` and `long` have the same size, but it will fail on platforms where `int` and `long` have different sizes.

Is this the sort of situation you are thinking about?

I think this is a flaw in the C language, as the aliases are not very opaque, but that's an issue for another day.

I suppose the "right" way to do this might look like

    typedef counter_t int;
    #define COUNTER_T_FORMAT "%d"
    // ...
    counter_t counter = 0;
    // ...
    printf( "The count = " COUNTER_T_FORMAT "\n", counter );

Then, if `counter_t` is updated, the `COUNTER_T_FORMAT` macro can be updated as well, and there is always a match between data type and format specifier, which should be portable across all compilers and platforms.

It should be noticed that C++ doesn't have this problem because of the way overload resolution works on `operator<<`.

winspool:
> If we add the macros defined by other compilers, it makes it appear that we are trying to support the extensions of those other compilers, which we aren't.
> So it is misleading to developers. 
> If feels like a hacky fix that might work in some simple cases, but not enough to be provide the sort of portability you are envisioning. 

I have a different opinion here

> A program that depends on gcc extensions, and uses gcc's macros to select for them, 
> is unlikely to compile with Open Watcom even if Open Watcom defined the same macros.

Using a different code path can also be required without using a compiler extension.
Example usage in OpenWatcom: compare filenames (ignoring case on Windows) 

> A an actual porting effort is going to be needed in any case,

That is your wrong opinion,
when the code is already ported for Linux and Windows,
but the port does not work with OpenWatcom, because the predefined macros are missing.

> and in that case, the code could be edited to use the Open Watcom macros.

And that us unneeded work, when OpenWatcom supports additional predefined macros

> At least, I think so?"

You thing so, but my opinion is different.

> On the other hand, I can understand the point you're making. 

Great. A step in the correct direction

> We do have `owcc` which provides a gcc-like interface to Open Watcom 
> to facilitate using Unix Makefiles out-of-the-box. 

And that's great.

> It could be said that providing gcc-like macros
> is just another example of that sort of thing.
> Even if it doesn't always work, it might work well enough, 
> enough of the time, to make it worthwhile.

Yes, please do it

> Yet, it could be argued that having `owcc` is a mistake 
> for the reasons I mentioned earlier.
> It misleads developers into thinking we are more compatible with gcc than we are.

Nope.

> How about this compromise position: 
> We change `owcc` so that when it launches the compilers it provides them with `-d` options to activate certain gcc-like macros. 
> The compiler remains true to itself, and the compatibility tool, which is known to be imperfect, even if sometimes useful, deals with the matter.

That's ok for me. As long as it works, the implementation is unrelated.
But i expect the code change to be more complex, than my proposed patch.


Argh...
I updated my branch, but github closed this Pull Request in my name.
That was not my whish.

winspool:
Synced Patch to current HEAD

ArmstrongJ:
I'm inclined to agree with @jmalak for the most part.  Following any sort of gcc-standard-esque behavior is never going to be a good idea.  Additionally, because there are no "standards" associated with, especially, architecture macros, OW will never keep everyone happy (have a look at [this mess](https://sourceforge.net/p/predef/wiki/Architectures/)...).

Being practical, however, I do think possibly adding _some_ macros to, specifically, owcc might be nice.  It seems like some of the dunder-wrapped macros are relatively common in my experience.  However, there being no standard makes this task a nightmare.  I would suggest that maybe if a macro appears in both clang and gcc, it might be okay to add, but clang has done a lot to copy gcc's behavior in this regard, so perhaps you'd need to look at a few others.  The task sounds miserable.  Dunder-wrapped target OS macros and maybe some arch stuff could be safe. 

Another thing to consider is that there are projects that are aware of and work properly with OW as-is.  These changes could quite possibly break them. 

I think in this particular patch, though, the linux case has to be reworked in regards to the QNX fall-through areas.  Those changes look especially dangerous, though I don't know if anyone is using OW on QNX any longer.  Best not to break it, though.

winspool:
> I think in this particular patch, though, the linux case has to be reworked in regards to the QNX fall-through areas. Those changes look especially dangerous, though I don't know if anyone is using OW on QNX any longer. Best not to break it, though.

QNX was alteady there ("__UNIX__") and i just reordered that part.
But that can easily adjusted.

jmalak:
I am happy for this discussion because it is important even if it looks as very simple non-dangerous change.
I understand what @winspool want but it must be clear what is possible and what not. Most people are now working on one main platform and sometimes the other that it is mostly covered by one compiler. They tend to accept what they use as standard and simply expect that all in world is same. OW works for many platforms (some is now exotic and not used) but we need to hold backward compatibility because we stay only one of a few compilers which support it and I think only one which is under active development.

I have a few note to owcc utility. 
- this utility was created as POSIX compatible cc utility for OW, that it can not be used for this solution
- it could be possible to create similar driver for compatibility with gcc which could solve such issues as translation gcc options and handle macros somehow, OW has similar drivers for MS commands (cl, lib, link, ..), but they are not up-to-date. Such gcc driver could be based on this or on owcc utility.

PR #1040, opened 2023-01-29
Add orl and nwlib mips code

lakor64:
The following PR should add orl and  nwlib for reading/writing coff code for Windows NT MIPS.
(Perhaps 0x8 was ok rather than 0xB actually)
last commits already implement this code 😅


lakor64:
last

PR #1050, opened 2023-02-19
* fix define _diskinfo_t in bios.h to _ibm_diskinfo_t

GateLinker:
`bios.h` did define `_diskinfo_t` as `_ibm_diskinfo`, which does not exist and I found its usage in an old codebase.
I think it should point to `_ibm_diskinfo_t`, so same as for `diskinfo_t`


jmalak:
I checked older Watcom versions and I think so it is a bug.

PR #1054, opened 2023-03-15
Add shell exports for CC and CXX, when building OpenWatcom

winspool:
Without this fix,
the Opwn Watcom build system does not use the selected compiler,
but fallback to c99 and c++ instead.

(gcc and clang are installed and clang is selected for OWTOOLS.
c99 and c++ are provided by gcc)

--
Regards ... Detlef

### DIFF
--- a/cmnvars.sh
+++ b/cmnvars.sh
@@ -30,9 +30,13 @@ if [ "$OWTOOLS" = "WATCOM" ]; then
     echo export OWTOOLSVER=__WATCOMC__>getversi.gc
     wcc386 -p getversi.gc >getversi.sh
 elif [ "$OWTOOLS" = "CLANG" ]; then
+    export CC=clang
+    export CXX=clang++
     echo export OWTOOLSVER=__clang_major__>getversi.gc
     clang -x c -E getversi.gc -o getversi.sh
 elif [ "$OWTOOLS" = "GCC" ]; then
+    export CC=gcc
+    export CXX=g++
     echo export OWTOOLSVER=__GNUC__>getversi.gc
     gcc -x c -E getversi.gc -o getversi.sh
 elif [ "$OWTOOLS" = "VISUALC" ]; then


jmalak:
I don't understand why it doesn't work for you.
Clang or GCC compilers are defined explicitly in build/mif/local.mif in dependency on OWTOOLS setup and it is used by all build commands. It cannot use system compiler (POSIX).
Only exception is for pre-bootstrap build wmake and builder tools where strict POSIX is used and therefore it use POSIX cc command.
How you check that wrong compiler is used?
To check set OWVERBOSE=1 environment variable to see exact command used by OW build system.

winspool:
> I don't understand why it doesn't work for you. 

You already explained the bug in this answer

> Clang or GCC compilers are defined explicitly in build/mif/local.mif in dependency on OWTOOLS setup and it is used by all build commands. 

That's ok

> It cannot use system compiler (POSIX).

That's wrong
 
> Only exception is for pre-bootstrap build wmake and builder tools where strict POSIX is used and therefore it use POSIX cc command.

And that's the bug in OW.

I set "OWTOOLS=CLANG", but the OW build system uses c99:
* that's a link to "/etc/alternatives/c99" and 
* that's a link to "/usr/bin/c99-gcc"
* and that's a script provided by gcc (which calls "gcc -std=c99")

With my patch, c99 is not used, but $CC instead 

> How you check that wrong compiler is used? 

I changed the "/usr/bin/c99-gcc" script and found that in build/binbuild/bootx.log

> To check set OWVERBOSE=1 environment variable to see exact command used by OW build system.

Build is running with "OWVERBOSE=1" to verify,
if the build breaks again with:

Error(F38): (<my_path>/bld/idedemo/src/os2/box.obj) does not exist and cannot be made from existing files

jmalak:
I am sorry, but if you don't use standard build environment don't expect that we will change OW build system for your own environment how you want.

**OW build system doesn't use CC or CXX environment variable anyway that this value is useless for OW build system.**

Try to look on our build server which use CLANG and GCC build how it works, there is no similar problem which you describe. [CI Build](https://dev.azure.com/open-watcom/open-watcom-v2/_build/results?buildId=8596&view=results)

jmalak:
You can setup CC or CXX environment how you want because OW build system doesn't change or use it any way.

You need only ensure that configured toolchain (OWTOOLS variable) is properly working for host system before you run OW build. OW doesn't change anything related to selected toolchain only set command line (select necessary tools options etc.)
If you somehow manipulate with gcc comand behaviour then it is your bussines not OW build system.

winspool:
> I am sorry, but if you don't use standard build environment

I'm sorry, but i use a very common development system,
and i expect, that the setup of the build system is special.

I installed my system recently on a bigger ssd and you can try it yourself
* Install ubuntu
* install 'build-essential' (includes gcc, g++, make, ...)
* Install clang

Now the system has gcc and clang installed.

I think, the build system is special, 
because it does not use 'build-essential', but clang and make.
In that case, the build system is special, because only one
compiler is installed.

My requirements for using multiple compiler on a development system (without the overhead of a VM or something like docker)
* Checking,  that the project is portable
* Checking for other warning messages


> don't expect that we will change OW build system for your own environment how you want.

I expect, that the build system get fixed, 
and for that, i invested time and prepared a pull request.

> Anyway OW build system is not Linux centric and it will not be anytime. 
> It is working on DOS, OS/2, Windows and Linux and it is toolchain centric because it needs support various target platforms.

I know. And that's also reason, why i am looking at OpenWatcom,
in addition of using gcc or tcc

> **OW build system doesn't use CC or CXX environment variable anyway that this value is useless for OW build system.**

Sorry, you are wrong here.
I collected this callchain for an OpenWatcom bootstrap:
* build.sh calls make for bld/wmake/posmake.
* the build rules in posmake use '$(CC)'
* without my fix, $(CC) is not defined
* the installed make (from build-essential) is gnu make
   in posix mode (mode selected by '.POSIX:' in posmake)
* without $(CC), gnu make is using the default posix c compiler:
   /usr/bin/c99
--> The OpenWatcom build system is using $(CC)

In addition, the OpenWatcom build system explicit sets
CC and CXX. 
Look at 'cmnvars.sh'. 
It's just a line below where the diff of my patch stopped.
(old: line 39+line40, with my fix: line 43 + line 44).

Since you did not see it, that my fix just extends 
the settings for 'CLANG' and 'GCC', similar to the settings
which are already present for 'VISUALC', 
I can assume, that you did not even looked
at the patched file after my fix.

I'm a bit sad, that you always deny fixes from new developer.

I really want to help to improve OpenWatcom, 
but learning to understand the huge codebase
and learning previous design decisions is difficult.

For this pull request, i expected an apply and not again a long discussion.

> Try to look on our build server which use CLANG and GCC build how it works, there is no similar problem which you describe.

See above. The build system is special, as it has only one compiler installed.

My pull request triggered a CI build, and i looked at the result page.
The diagram was there, but i could not open any result.
Azure reported an internal error, when i looked at, so i read 'azure.yml' and the 'ci' directory in the OpenWatcom tree, 
but i don't understand that yet.

> [CI Build](https://dev.azure.com/open-watcom/open-watcom-v2/_build/results?buildId=8596&view=results)

Yes, i saw a similar page in the CI build after my pull request.
Since Azure is working again, I will look at your link, when i'm back from work.

-- 
Regards ... Detlef

jmalak:
First of all you should read POSIX standard make specification https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html
any POSIX compliant make must define CC environment variable to point to C compiler.
You have no problem with building wmake by make because it relates to CC environment variable and run proper compiler.
If it is not working then it is your box problem or your installed make is not POSIX compliant.
all other build is done by wmake which doesn't use POSIX definition but wmake makefiles which are independent on POSIX environment.
Therefore after build wmake nothing is dependent on anything related to host system.
only wmake with make files are used to build all other.
As second is build builder tool. Builder tool process *.ctl files which handle full OW build together with wmake, only these two tools are responsible for handlig full OW build on any supported host and for any target system.

There is nothing magical simply two clever tools and two types of configuration files (*.ctl and make files) which traverse OW source tree and build what is necessary.

I don't want lost my time with this discussion, which is absolutely useless.
Only thing related to POSIX CC is build of wmake as first step, nothing more.
if something is not working then it is bug in some code which can not be compiled or in configuration files or in wrong build environment (it is most probable).

winspool:
> First of all you should read POSIX standard make specification https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html

Thanks for the link

> any POSIX compliant make must define CC environment variable to point to C compiler.

From a search on the website, the first location of CC found
is the definition of the default value for CC: c99.
I already wrote, that gnu make in POSIX mode uses c99,
when the caller of gnu make did not set CC.
So gnu make is correct to use c99.
That is independent from the choice of the value of OWTOOLS
in the OpenWatcom build system.

My fix makes sure, that a selection of OWTOOLS
is respected also during the OpenWatcom bootstrap by make.


> You have no problem with building wmake by make because it relates to CC environment variable and run proper compiler.

Yes, and without my fix, make is using the default, which is c99
and on any default ubuntu system which used the
suggested package 'build-essential'  this is gcc.
And without my fix, this is not the compiler, 
which i forced to use by setting OWTOOLS to CLANG

> If it is not working then it is your box problem or your installed make is not POSIX compliant.

Both is wrong.
wmake compile works and make is posix compatibe.

But the compiler used for compiling wmake
is not the compiler toolchain, 
which I as the builder of OpenWatcom selected 
by setting OWTOOLS  t CLANG

That is the bug in the OpenWatcom build system,
and to fix this bug in the OpenWatcom build system,
i invested my time and i created a pull request.


> all other build is done by wmake which doesn't use POSIX definition but wmake makefiles which are independent on POSIX environment.
> Therefore after build wmake nothing is dependent on anything related to host system.
> only wmake with make files are used to build all other.
> As second is build builder tool. Builder tool process *.ctl files which handle full OW build together with wmake, only these two tools are re


Every sentence here is correct.
I learned it by using it multiple times, reading the OW wiki,
the log files and the source.


But that has nothing to do with bug in the OpenWatcom build system, that a different compiler is used, when i select
OWTOOLS=CLANG


With your style of deny all changes to your baby from other persons, 
it is very resonable, why no external developer stays at
OpenWatcom and wants to invest time to bring OpenWatcom forward.

Let me compare that to Wine,  which is not perfect, but Wine
* has a default setting of win10 (and can be set to Win11)
* has 16bit, 32bit and 64bit support, 
* runs on Linux, various *BSD, MacOS 
* runs also on Chromebock/Android
* supports most old programs and games
* supports many new Programs and Games
* supports also newer Windows programs/Games requiring 
   modern API, DirectX up to 12, and other modern things
* supports some strong copy protections/
   system checks included in new games
* was picked up and is still in use by other companies
   (Many very well known programs used/ are using Wine.
     Picassa, Steam on Linux,  Teamviewer, many others...)

With your style of leading a project, Wine would be stuck,
where Wine was, when the first company left Wine.
That was Corel for Corel Draw on Linux
and the development Level was 16bit Windows (Win 3.1/3.11)
with a few parts in 32 bit for Windows 95.

Sorry for that, but that is my feeling with your style 
of deny everything with useless long arguments, 
instead of open your mind.
Please think about the reported probems of users of your baby,
accept, that there are bugs,
which you do not know and which you think,
that this issues are always failures from the user on the user system.
Take it easy, thank others, 
who try to help moving your Baby OpenWatcom forward
and accept their help.

######

Same issue recently was about gcc is using __i386__ 
and OS defines, which i fixed and created a pull request,
but you did not liked that.
OpenWatcom still does not work for many unix software,
which works fine with gcc.
On the other side, OpenWatcom supports many
extensions from the Microsoft compiler on dos and windows.

I think, this was a very stupid decision against unix and all modern compiler (not only gcc, but all big compiler).

With your decision against supporting compiler specific  extensions, 
why do you not mark the Microsoft extensions of C
which are implemented in OpenWatcom as deprecated?


For your Info:
clang is compatible to gcc with this defines
and clang is integrated in the compiler suites by
Microsoft, Intel, Embarcadero (__BORLANDC__).
The EDG compiler suite (and some other compiler) are using gcc.
That is something named marked dominance.

Where is OpenWatcom?
No longer well known, not even mentioned on cppreference.
The nice for OpenWatcom are the old DOS programs.

Is OW still in use for other dead systems?
Novell Netware, NT4/mips, NT4/alpha, PC98, 
other things from the DOS timeframe? 
Or for something from the OS/2 timeframe?

########

Another example is my investment to test yarpgen with owcc.
Try yarpgen (v1) for seed 7124 compared to seed 25.

tcc (64 bit mode) works, but the needed Compile time
increased from 0.4sec to >12sec (factor ~30)

Compile time for clang (64 bit mode) increases from
~25 sec to ~15min  (factor ~30)
(just at the memory limit of my 12GB laptop)

Compiling with gcc (64bit mode)runs in the memory limit of my laptop and gcc swapped for 79min

All three compiler produced the same result.
(Same results for tcc in 32bit mode and tcc in win32 mode)

I did not try owcc yet.
With the 30x time increase factor for tcc and clang 
compared to  yarpgen (v1) for seed 25 
and the 500sec needed by OpenWatcom for seed 25,
I expect the time needed for OpenWatcom
with the active bugs/unimplemented features in OW 
to compile this singe testfile around 4h plus swap time.
And with the other bugs/unimplemented features in OpenWatcom, 
the chance to get a correct result is low.

####

To repeat myself:
i want to help OpenWatcom, 
but when everything is denied, i'm sad.


-- 
Regards ... Detlef

jmalak:
Sorry, I wrote you several times that make is not used by OW build system at all.
It is used only for building wmake (bootstrap version), nothing more.
Why you talk about this all the time again and again if building of wmake is without problem?
It is absolutely useless discussion. 
I don't believe that GNU make is not POSIX compliant if it is setup to POSIX mode, see first line of posmake make file.
I didn't have any trouble with it. Only OSX build in the past have some trouble but by use of strict POSIX mode fix it.

As example this is absolutely useless change and discussion which only consume time for nothing.
There is lot of reported problems to fix, nobody want to fix anything.
I am able to help anybody if I know, but I don't want work on something not useful.

jmalak:
Anyway to your problem with idedemo project build and /bld/idedemo/src/os2/box.obj build.
This project is specific, because it pre-build objects necessary for demo example described in OW documentation.
It is defined as IDE project.
for build it following  steps are necessary:
- ide2make tool must be prebuild (by bootstrap phase)
- ide2make tool generate wmake make files from IDE project configuration file
-  IDE project use all path related to OW installation that for pre-build is necessary to change all path in make files to proper path of OW build system. For it sed tools is used and it modify make files generated by ide2make tools to make file capable to run in OW build system. there is appropriate sed scripts which replace all necesary info in make files
- build object for each target system

It is relatively complex project, that is necessary check which step is casing a problem.
Simple is to go into bld/idedemo folder and run **`builder OWVERBOSE=1`** command, it dispaly details abot all commands used for building these object file.

PR #1057, opened 2023-03-17
Fixed a problem that made OW's standard library not compliant with the c++ standard.

Count-MHM:
`ios::in` and `ios::out` should always be set (even if not explicitly set in arguments) for `ifstream` and `ofstream`.
Constructors handled that correctly, but `open` function didn't.

### DIFF
--- a/ci/ow/h/fstream
+++ b/ci/ow/h/fstream
@@ -169,7 +169,7 @@ namespace std {
 #pragma pack( __pop )
 
   inline void ifstream::open( char const *__n, ios::openmode __m, int __p ) {
-    fstreambase::open( __n, __m, __p );
+    fstreambase::open( __n, __m | ios::in, __p );
   }
 
   // **************************** OFSTREAM ***********************************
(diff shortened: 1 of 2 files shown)


jmalak:
@pchapin could you check this PR.

jmalak:
Peter thanks for review. You are right this is incorrect.
**`ci`** subfolder contains read-only things which are used by Azure Pipelines builds and is also used for some hosts for build OW and must not be changed. Here is copy of OW 1.9, DOSBOX, GhostScript etc.
Only **`bld`** subfolder contains OW sources.
As you noted it must be changed in **`bld/hdr/watcom`**

jmalak:
@Count-MHM Please correct it as Peter noted.

Count-MHM:
@Count-MHM @pchapin Corrected my mistake.

Count-MHM:
@jmalak @pchapin Corrected my mistake.

jmalak:
Please remove wrong change in **`ci`** subfolder.

Count-MHM:
@jmalak Done!

PR #1061, opened 2023-03-20
owcc: Handle more architectures: x64(x86_64,amd64), arm and a64(aarch64)

winspool:
Documentation and "-?" output are unchanged, as the tools are still not present.

When running on a x86_64 (x64) system, the default is still to use the 32 bit tools

Fist patch, to verify, that this direction is ok.

--
Regards ... Detlef

### DIFF
--- a/bld/wcl/c/owcc.c
+++ b/bld/wcl/c/owcc.c
@@ -92,6 +92,27 @@
 #define ASM         BPRFX "wasmps"      /* Open Watcom assembler             */
 #define _TARGET_    "MIPS"
 #define ARCH        TARGET_ARCH_MIPS
+#elif defined( _M_X64_AS_DEFAULT )
+#define CC          BPRFX "wccx64"      /* Open Watcom C compiler (64-bit)   */
+#define CCXX        BPRFX "wppx64"      /* Open Watcom C++ compiler (64-bit) */
+#define FC          BPRFX "wfcx64"      /* Open Watcom F77 compiler (64-bit) */
+#define ASM         BPRFX "wasm"        /* Open Watcom assembler             */
+#define _TARGET_    "x86 64-bit"
+#define ARCH        TARGET_ARCH_X64
+#elif defined( _M_ARM )
+#define CC          BPRFX "wccarm"      /* Open Watcom C compiler (32-bit)   */
+#define CCXX        BPRFX "wpparm"      /* Open Watcom C++ compiler (32-bit) */
+#define FC          BPRFX "wfcarm"      /* Open Watcom F77 compiler (32-bit) */
+#define ASM         BPRFX "wasarm"      /* Open Watcom assembler             */
+#define _TARGET_    "ARM 32-bit"
+#define ARCH        TARGET_ARCH_ARM
+#elif defined( _M_ARM64 )
+#define CC          BPRFX "wcca64"      /* Open Watcom C compiler (64-bit)   */
+#define CCXX        BPRFX "wppa64"      /* Open Watcom C++ compiler (64-bit) */
+#define FC          BPRFX "wfca64"      /* Open Watcom F77 compiler (64-bit) */
+#define ASM         BPRFX "wasa64"      /* Open Watcom assembler             */
+#define _TARGET_    "ARM 64-bit"
+#define ARCH        TARGET_ARCH_ARM64
 #else
 #define CC          BPRFX "wcc386"      /* Open Watcom C compiler (32-bit)   */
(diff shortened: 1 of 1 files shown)


jmalak:
What is purpose of such changes?

winspool:
Since i have a lot more time starting at April,
i want to do useful things for OW.

Getting a working OW on my Raspberry Pi would be great
and the owcc command line options are far easier for me
compared to calling wclxxx or wccxxx/wasxxx/wl directly.

But i did not tested/deceided yet,
which arch or which tool is easier for me:
aarch64 or arm and 
then wccxxx, wasxxx or wl together with ORL/OWL.

And i have to learn more about the tests in OW.

-- 
Regards ... Detlef

jmalak:
I think you don't understand how OW build works and what is necessary to deploy new architecture.
What you are trying to do is something as Russian Potemkin village when only rear wall is built on single street.

First of all, OW support only following architectures/targets:
DOS 16 and 32 bit
OS/2 16 and 32 bit
Windows 16 and 32 bit
Linux 32 bit
Netware 32 bit
RDOS 16 and 32 bit
QNX is not supported even if stuff exist, because it requires 3-rd party Software which is Licensed, therefore it cannot be build and test. It is used as template for some Linux stuff.
There is experimental support not fully implemented for 32-bit Windows Alpha architecture, Power PC and MIPS and for 32-bit Linux Power PC and MIPS.

No other platforms.

Therefore what you are trying to do is absolutely nonsense.
To have sense you must first create Code Generator and  port C run-time Library to such architectures.
Next modify all main tools to support this non-existent architectures, but without previous things you can not.

Anyway OW build system support OW build on various platforms, but it has nothing to do with creating executable for these platforms. It enables cross-compile from these hosts systems to supported targets, but not for host OS.
Now only checked OW build is for OSX, but it doesn't mean that you can use OW for OSX.
There are other possible (experimental) OW build platform as Linux ARM 32 and 64-bit and OSX ARM 32 and 64 bit but not supported for reason that we have no build/test platform.

jmalak:
If you want to use OW on Rapsbery-PI for cross-compile then you can build it with existing OW build system.
if you run in **`bld`** subfolder command **`builder rel os_linux cpu_arm .and .or`**

jmalak:
Anyway if you want to do something useful for OW try to look on existing bug reports and try to fix some of them.
If anything not clear or you need help then open issue on Github to get appropriate info/help.
OW build system is configured for many target even if it is not use regularly and not suitable for common user.
It is for developer of for people which understand it. Until we get some possibility to use new platform we can not support more.
Now we use Azure pipelines which handle only Intel Linux, Windows and Intel OSX.

winspool:
> I think you don't understand how OW build works 
> and what is necessary to deploy new architecture.

I had a generic knowledge, how OW works,
and every day, when i read through the source, 
or try something, i know much more.
But there are still many things, which i do not know.

As an example for learning,
i tried to build as much as possible of OW
with only a C compiler.
(commenting out the projects, which need a C++ compiler
or Assembler for the DOS extenders).

Trying to build OW with the tools created during bootstrap
would be another experiment in the future.

> First of all, OW support only following architectures/targets: 
...
> No other platforms.

Yes, at the moment. I know that.

> Therefore what you are trying to do is absolutely nonsense.

Are you afraid of adding a new architecture?

> To have sense you must first create Code Generator 
> and port C run-time Library to such architectures. 

And OWL + ORL + WAS and all other tools.

I also know, that the output for x86 is different.
So extending that for x86_64 is another big project.
But there is a fork of wasm/wlink, which was extended
to support x86_64 and newer ASM instruction sets (upto AVX):
https://www.japheth.de/JWasm.html
https://www.japheth.de/JWlink.html
Released under the Sybase Open Watcom Public License.


> Next modify all main tools to support this non-existent architectures,
> but without previous things you can not.

Yes, a big bunch of work.

> Anyway OW build system support OW build on various platforms, 
> but it has nothing to do with creating executable for these platforms. 
> It enables cross-compile from these hosts systems to supported targets, 
> but not for host OS.

Yes, I know, that compiling for aarch64 on aarch64 does not work yet.

-- 
Regards ... Detlef

jmalak:
It has no sense to develop front-end application if you have no core.
Much useful is to start with porting of OW C run-time library extension in bld/watcom/c/clibext.c to new platform.
It ensures OW will be compilable on these platforms by native bootstrap compiler if not yet implemented.
Next to fix all issues for porting to new platform.
Next can be develop support for new platforms to ORL, Librarian and Linker.

PR #1063, opened 2023-03-24
Make sure graphics state is inited before returning cursor shape

artlogic:
`_settextcursor` is documented to return the previous cursor shape before setting the new one - however if `_InitState` has not been called the previous cursor shape is undefined.


jmalak:
Thanks you for your contribution.
Submitted.

PR #1084, opened 2023-05-11
owcc: Fix handling of equal character with option -b

winspool:
Example:
owcc -b=dos -v -std=c99 hello.c -o hellodos.exe

The patch avoids calling wcc386 with the broken arg "-bt==dos" 
followed by wlink stop the compilation:
Warning! W1107: file __owcc00.lnk: line(2): undefined system name: system
Error! E3033: file __owcc00.lnk: line(2): directive error near 'system'
Error: Linker returned a bad status

--
Regards Detlef

### DIFF
--- a/bld/wcl/c/owcc.c
+++ b/bld/wcl/c/owcc.c
@@ -918,9 +918,9 @@ static  int  ParseArgs( int argc, char **argv )
         case 'b':
             Flags.link_for_sys = true;
             MemFree( SystemName );
-            SystemName = MemStrDup( Word );
-            /* if Word found in specs.owc, add options from there: */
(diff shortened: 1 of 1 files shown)


jmalak:
It is not bug in owcc, but bug is in help text.
The same issue is for option -o.
Correct help text for these options should be
-b <target>
-o <file>

jmalak:
> It is not bug in owcc, but bug is in help text. The same issue is for option -o. Correct help text for these options should be
> 
> ```
> -b <target>
> -o <file>
> ```

I fixed help text just now.

winspool:
Issue resolved by changing the documentation.

PR #1086, opened 2023-05-13
owcc: Accept more filename extensions as fortran source:.f77, .f90, .f95

winspool:
This is enough to build 3 of the 17 test from
The "Polyhedron Fortran Benchmarks" Suite
which is available from fortran.uk ( www.fortran.uk/pb11.zip )

These files compile with OpenWatcom:
air.f90, doduc.f90, linpk.f90

--
Regards ... Detlef

### DIFF
--- a/bld/wcl/c/owcc.c
+++ b/bld/wcl/c/owcc.c
@@ -115,7 +115,9 @@
 #define MAX_CC_OPTS 256
 
 #define IS_ASM(x)   (x[0] == '.' && (fname_cmp(x + 1, ASM_EXT) == 0 || stricmp(x + 1, ASMS_EXT) == 0))
-#define IS_FOR(x)   (x[0] == '.' && (fname_cmp(x + 1, "f") == 0 || stricmp(x + 1, "for") == 0 || fname_cmp(x + 1, "ftn") == 0))
+#define IS_FOR(x)   (x[0] == '.' && (fname_cmp(x + 1, "f") == 0 || stricmp(x + 1, "for") == 0 || \
+                    fname_cmp(x + 1, "ftn") == 0 || stricmp(x + 1, "f77") == 0 || \
+                    stricmp(x + 1, "f90") == 0 || stricmp(x + 1, "f95") == 0))
 #define IS_LIB(x)   (HasFileExtension( x, LIB_EXT ) || HasFileExtension( x, LIB_EXT_SECONDARY ))
 
 typedef enum {


jmalak:
I am not sure if it is right way how to define new extensions for file type.
Probably better solution should be to externalize extensions list for various file type in specs.owc file.
It enables user configuration in case that use prefere other extensions.
There is now 4 file types
**`assembler source`**
**`C source`**
**`C++ source`**
**`Fortran source`**
we could add
**`object files`**
and
**`library files`**

pchapin:
Is it appropriate for Open Watcom Fortran to accept the .F90 extension? It isn't an F90 compiler, after all. The .F77 extension would make sense, though. Perhaps I misunderstand the PR.

jmalak:
It is good note. OW Fortran compiler use 1977 standard that f90 or f95 extension could be confusing.
If we externalize list of extensions for Fortran source file then we will add only f77 extension and others could be add by users as customization.

winspool:
> Is it appropriate for Open Watcom Fortran to accept the .F90 extension?

Yes!

The old file extensions (examples: ".f", ".for", ".ftn") are widely used
for Fortran files in the fixed style format
while ".f90" is used for Free style format.

https://fortranwiki.org/fortran/show/File+extensions
https://fortranwiki.org/fortran/show/Fortran+90
https://riptutorial.com/fortran/example/31322/extensions-and-meanings

Some people thought that the file extension is only related to the
needed / used Fortran version and use ".f03", ".f08" of ".f15" 
for more modern Fortran sources, but they are rare since these 
extensions are not supported by all relevant fortran compilers.

( OpenWatcom is not relevant here and even marked as "Discontinued":
https://fortran-lang.org/en/compilers/ ) 

Today, ".f90" is only a hint for free style format,
but a specific file extension is not required to distinglush
fixed style / free style.

OpenWatcom supports free style Fortran sources.
At least, some benchmarks are Fortran F77, but Free Style
and the tree has "Free format input":  
https://github.com/open-watcom/open-watcom-v2/blob/master/bld/f77/f77lib/c/freein.c

> It isn't an F90 compiler, after all. 

I did not test, if 100% of the Fortran 90 features work,
but that doesn't matter. 
Free Format works, and that is the most important F90 feature.

> The .F77 extension would make sense, though. 

gfortran uses uppercase filename extensions as a marker to use the preprocessor before calling the compiler.
OpenWatcom does not support that, but i do not care.

I just found a reference to the testsuite and tried it.
Fixing file extension support in owcc looked like low hanging fruit,
but i have to explain again a simple fix with a long text.

> Perhaps I misunderstand the PR.

Yes. Real Wold Free style Fortran source files use ".f90"
and OpenWatcom works at least for some of the benchmark examples.

OTOH, OpenWatcom is a dead dinosaur and almost nobody cares about OW.
(Only a few people here.)
As example, where is OpenWatcom listed / mentioned
in any recent compiler comparsions?
Fortran:
https://polyhedron.com/?page_id=175
C++ (Lowest version listed is C++11, so OW is not mentioned here):
https://en.cppreference.com/w/cpp/compiler_support
https://en.cppreference.com/w/cpp/compiler_support/11

C: (OW is not even listed as a C99 Compiler)
https://en.cppreference.com/w/c/compiler_support


LLVM gets a new Fortran frontend which is support by big companies
(IBM, probably more)


-- 
Regards ... Detlef

winspool:
> I am not sure if it is right way how to define new extensions for file types.

It is the easiest way, and i do not see any need, to create a framework
to add more file extensions.

> Probably better solution should be to externalize extensions list for various file type in specs.owc file.

Do you really think, that there is a need to add so many new file extensions
in the near future, that creating, fixing, testing and maintaining
a new framework for file extensions is faster, less work and less code  
compared to the change in my pull request?

> It enables user configuration in case that use prefere other extensions.

That's far too complicated.

From your discription, i think that you define a user as someone,
who configures and builds OpenWatcom

I define a user as someone, who installs OpenWatcom and uses it
to compile existing sources / building existing software projects.

So everything, that does not work "out of the box" with OpenWatcom
is the same as everything, that is not available in OpenWatcom.


> There is now 4 file types
> **`assembler source`**
> **`C source`**
> **`C++ source`**
> **`Fortran source`**
> we could add
> **`object files`**
> and
> **`library files`**

I have no idea, what do you describe here.
Is it something already present in the source?
(owcc as recent example).

Can you give me a hint?

-- 
Regards ... Detlef

winspool:
> It is good note. OW Fortran compiler use 1977 standard that f90 or f95 extension could be confusing.

See my other answer:
".f90" as a hint for Free Form Fortran

When you strictly do not like ".f95", that skip that part,
or let me know, then i will spend more time and resend it.

Oh, i did not looked at any documentation change yet.
I already need too much time to explain the OpenWatcom Team
even the simplest bug fixes
and did not changed my recently installed system yet
to build the OpenWatcom documentation.

> If we externalize list of extensions for source file types

Not needed and too complicated.

> add only f77 extension and others could be add by users as customization.

Do you really have so many free time to spend on a new 
Framework for file extensions for every tool and the examples
in the whole OpenWatcom tree?

pchapin:
@winspool Are you sure Open Watcom Fortran accepts free formatted input? Consider this program:

    PROGRAM Hello
    REAL A, B, C
    REAL d, r1, r2

    READ *, A, B, C
    d = B * B - 4 * A * C
    IF ( d .LT. 0 ) THEN
      PRINT *, 'Complex Roots'
    ELSE
       r1 = ( -B + SQRT( d ) )/( 2 * A)
       r2 = ( -B - SQRT( d ) )/( 2 * A)
       PRINT *, r1, r2
    ENDIF
    END

The significance of the listing is that (most of) the lines start in column 1. Open Watcom produces the following errors:

    hello. For(1): *ERR* CC-02 invalid character in statement number columns
    hello.for(2): *ERR* CC-02 invalid character in statement number columns
    hello.for(3): *ERR* CC-02 invalid character in statement number columns
    hello.for(5): *ERR* CC-02 invalid character in statement number columns
    hello.for(7): *ERR* CC-02 invalid character in statement number columns
    hello.for(8): *ERR* CC-02 invalid character in statement number columns
    hello.for(9): *ERR* CC-02 invalid character in statement number columns
    hello.for(10): *ERR* CC-02 invalid character in statement number columns
    hello.for(11): *ERR* CC-02 invalid character in statement number columns
    hello.for(12): *ERR* CC-02 invalid character in statement number columns
    hello.for(13): *ERR* CC-02 invalid character in statement number columns
    hello.for(14): *ERR* CC-02 invalid character in statement number columns
    hello. For(14): *ERR* EN-01 missing END statement

My understanding of free format is that the column where code begins is no longer relevant. In contrast `gfortran` *does* compile the example when it is stored in a .f90 file. Open Watcom behaves the same (i.e., produces the same errors) when the program is in a .f90 file.

Peter

ArmstrongJ:
I feel like I'm somewhat knowledgeable in the realm of Fortran standards since I deal with that language daily.  Open Watcom absolutely does _not_ accept free-format Fortran, introduced in the Fortran 90 standard.  The **.f90** extension (as well as **.f95**, **.f03**, **.f08**, etc.) imply free-format Fortran code.  The Fortran standard actually has no mention of file extensions, but using **.f90** for free-format code that conforms to *at least* the Fortran 90 standard is generally accepted.

I fear that saying "generally accepted," though, is too soft a statement.  It would be irresponsible and incorrect for Open Watcom to accept those later file extensions.  It would be the same as Open Watcom's C compiler accepting **.m** extensions because an objective-C file *might* compile if it doesn't use those objective-C features that make it objective-C.  

Almost all fixed-format Fortran uses **.f** and **.for** extensions.  I have occasionally seen oddball source files that use **.f77**, **.ftn**, and **.ftn77**, but they are somewhat rare and not generally accepted these days.  I would argue leaving **owcc** as is would be fine.  Accepting **.f90** would simply be wrong.

winspool:
> Are you sure Open Watcom Fortran accepts free formatted input?

I assumed that, because OpenWatcom was able to compile some
examples of the Fortran benchmarks with ".f90" extensions
and the OpenWatcom source has source file 
with the description "Free format input":
https://github.com/open-watcom/open-watcom-v2/blob/master/bld/f77/f77lib/c/freein.c

> Consider this program:
behavior of 
Sorry, my Fortran knowledge is very rare.

>  In contrast gfortran does compile the example when it is stored in a .f90 file.

Some of my other patches where rejected, because
"OpenWatcom does not clone other compiler, but only follows the standard",
so that "gfortran" can compile your example is only a nice info,
but no reference.

Sorry, but currently i have no time for a deep Fortran study

-- 
Regards ... Detlef

winspool:
Hi Jeff, thanks for sharing your knowledge.

> Open Watcom absolutely does _not_ accept free-format Fortran, introduced in the Fortran 90 standard.

Oh, then i misinterpreted, that OpenWatcom is able to compile
some of the Fortran benchmarks which have the ".f90" filename ending
together with the "Free form input" source file at
https://github.com/open-watcom/open-watcom-v2/blob/master/bld/f77/f77lib/c/freein.c

> I would argue leaving **owcc** as is would be fine.

Without my patch, owcc starts the c compiler 
and fails in 17 out of 17 of the Fortran benchmarks,
but with my patch, the Fortran compiler is used for the ".f90" files
and 3 out of 17 benchmarks works.
And another benchmark compiles, but linking fails (CPU_TIME missing).

-- 
Regards ... Detlef

winspool:
> If you look into wcl/owcc code there are check macros as IS_ASM, IS_OBJ, IS_FOR, ....
> There is also function SrcName, which identify file type used by owcc.

Ah, you mean these things in owcc

> They qualify type of files that owcc is able to add it to appropriate tool command line.

Yes, i extended that, so owcc does not call the C compiler,
but the Fortran compiler on the Fortran ".f90" benchmark source files

> What I means is to add some new items to specs.owc file to define appropriate types extensions there and completely remove hard coding of extensions from source code.

I am still protesting against breaking file extension handling in such a way.

> It is not too complicated, because handling of specs.owc file is in code and most of type identification code is there.

Wrong
It's complicated, because "specs.owc" is currently only used in in owcc.

You forgot the ide, the debugger, and all other tools, which handle object files and lib files just fine without a new framework.
When you add a configuration, it has to be consistent everywhere.

Do you really think, that every tool should get a new dependency on "specs.owc"?
Please do not add such a bloat.

> Review of code is required to locate all places where changes are required.

jmalak:
> > If you look into wcl/owcc code there are check macros as IS_ASM, IS_OBJ, IS_FOR, ....
> > There is also function SrcName, which identify file type used by owcc.
> 
> Ah, you mean these things in owcc
> 
> > They qualify type of files that owcc is able to add it to appropriate tool command line.
> 
> Yes, i extended that, so owcc does not call the C compiler, but the Fortran compiler on the Fortran ".f90" benchmark source files
> 
> > What I means is to add some new items to specs.owc file to define appropriate types extensions there and completely remove hard coding of extensions from source code.
> 
> I am still protesting against breaking file extension handling in such a way.
> 
> > It is not too complicated, because handling of specs.owc file is in code and most of type identification code is there.
> 
> Wrong It's complicated, because "specs.owc" is currently only used in in owcc.
> 
> You forgot the ide, the debugger, and all other tools, which handle object files and lib files just fine without a new framework. When you add a configuration, it has to be consistent everywhere.
> 
> Do you really think, that every tool should get a new dependency on "specs.owc"? Please do not add such a bloat.
> 
> > Review of code is required to locate all places where changes are required.

Sorry, I don't understand what you wan't to fix.
What have owcc utility to do with IDE?
Do you understand how IDE works?

jmalak:
I agree, I will close this issue.
Basic idea for this PR is bad.
I add new issue #1088 with enhancement request for owcc to externalize file type extensions, that can be modified (can be useful).

PR #1092, opened 2023-05-22
owcc: extend -mcmodel argument handling

winspool:
I updated the patch:

owcc: accept also long names for the -mcmodel option

In addition to the short memory model names (t, s, c, m, l, h),
owcc accepts now also long memory model names as used by gcc
(tiny, small, compact, medium, large, huge)

--
Regards ... Detlef

### DIFF
--- a/bld/wcl/c/owcc.c
+++ b/bld/wcl/c/owcc.c
@@ -741,7 +741,9 @@ static  int  ParseArgs( int argc, char **argv )
             wcc_option = false;
             break;
         case 'm':
-            if( ( strncmp( "cmodel=", Word, 7 ) == 0 ) && ( Word[8] == '\0' ) ) {
+            /* accept short mcmodel names (ow style) and long names (gcc style) */
+            /* examples: -mcmodel=h and -mcmodel=huge */
+            if( ( strncmp( "cmodel=", Word, 7 ) == 0 ) && ( Word[7] != '\0' ) ) {
                 if( Word[7] == 't' ) {      /* tiny model */
                     Word[0] = 's';          /* change to small */
                     Flags.tiny_model = true;


jmalak:
For me it works.
command 
**`owcc -mcmodel=h -v blb.c`**
gives you
**`wcc386 -mh -fo=.obj -fr blb.c`**

Help display syntax for **-mcmodel**
```
-mcmodel={c,f,h,l,m,s,t}     memory model
  c                          - memory model compact
  f                          - memory model flat
  h                          - memory model huge
  l                          - memory model large
  m                          - memory model medium
  s                          - memory model small
  t                          - memory model tiny
```

jmalak:
I checked it with OW 1.9 and there is bug
**owcc -mcmodel=h blb.c**
command doesn't put -mh option to compiler command line
but syntax is same
**-mcmodel={f,t,s,m,c,l,h}        memory model**

winspool:
> I checked it with OW 1.9 and there is bug
> neither
> **owcc -mcmodel=h blb.c**
> nor
> **owcc -mcmemory=huge a1.c**
> command put -mh option to compiler command line
> but syntax is same as for OW 2.0
> **-mcmodel={f,t,s,m,c,l,h}        memory model**


I used, what i know since years from gcc
for windows and dos, and what i used recently for aarch64: 
the long name  "-mcmodel=huge" 
but that failed with owcc.

I updated the patch to accept 
long memory model names
in addition to the short memory model names.

When that enhancement is acceptable,
i prepare a change for the documentation and the owcc help

-- 
Regards ... Detlef

jmalak:
I am sorry owcc tool is POSIX compliant.
Where is defined such options?
You don't fix any error, but introduce redundant form.
what is purpose for this?

winspool:
> I am sorry owcc tool is POSIX compliant.
> Where is defined such options?

Nowhere inside POSIX.

The leading compiler defines the command line arguments,
and for unix systems, gcc was the first leader.

For many projects, a "CC=owcc" or "CXX=owcc" would be enough,
when the basic command line args are the same between gcc and owcc.

> You don't fix any error, but introduce redundant form.
> what is purpose for this?

command line compatibility with gcc, so owcc can work as a dropin replacement
without changing a lot in every makefiles.

That was also done by clang/llvm, to make this newer toolchain a simple dropin for gcc.


And that was already done inside OpenWatcom with far more code
for the leading compiler used in DOS/Windows: 
The MS command line tools in bld/mstools

When you still think, that command line compatibility is a redundant feature, 
then remove the redundant code inside bld/mstools.

-- 
Regards ... Detlef

PR #1094, opened 2023-06-03
Add missing newline, fix a copy&paste error

winspool:
I started with objread from bld/orl/test,
but that segfaults on ELF64.

A git grep for ELFOSABI made the copy&paste failure visible

wdump looked more advanced on the first sight (because it does not segfaults),
but that assumption was wrong: ELF64 does not work there either.
Elf64_Ehdr is present in bld/watcom/h, but unused in all projects other that bld/orl.

-- 
Regards ... Detlef


jmalak:
Thanks for your contribution.

PR #1098, opened 2023-06-14
owcc: Allow more values for the C and C++ standard

winspool:
The C and C++ compiler can react on more standards, so owcc should allow them also

I did not change the online help for "-std=..."  yet:
-std=wc is used in the owcc command line help ( owccopts.gml )
-std=ow is used in the source file

Which one is correct?

--
Regards ... Detlef


jmalak:
what is purpose of such change if OW compilers doesn't support such options any way?

winspool:
> what is purpose of such change 

This is a bugfix.

> if OW compilers doesn't support such options any way?

The do support that.
Features allowed at the commandline of wcc* (-za99 / -za23) or wpp* (-za0x)
are not possible when using owcc.
(They are not documented on the command line help of wcc* and wpp*) 

For my project, i tried OpenWatcom with autotools,
but autotools does not detect OpenWatcom as a C++98 compiler.

Before sending a patch to set "__cplusplus" to 199711L (as required by C++98,
and which is the first thing checked by autotools to detect the suppoeted c++ standard),
i stumbled on "enable_std0x" in bld/plusplus and send this bugfix first.

My bugfix also accepts "-std=c23" and forwards that as "-za23" to wcc*. 
You added "-za23" to the C Compiler, but you forgot the related update to owcc.

-- 
Regards ... Detlef

winspool:
> Anyway you are right, there is mistake in run-time help for "std" option in owccopts.gml that it should be fixed to -std={c89,c90,ow}, but in OW on-line help documentation it is correct.

I want to get my bugfix for "-std=*" first to be accepted,
because updating the owcc commandline help changes the same line again

jmalak:
> My bugfix also accepts "-std=c23" and forwards that as "-za23" to wcc*,
> because you added "-za23" to the C Compiler, but you forgot the related update of owcc.

No, this option is WIP therefore it is not documented etc.

pchapin:
It makes sense to me to change -za0x to something else. That name was chosen in the long-ago days before it was known when C++ 2011 would be finished. The standard was called "C++ 0x" for a long time during its development. Obviously, that is all ancient history.

An obvious change would be to rename the option za11. The problem, though, is that such a name is ambiguous. Are we talking about C 2011 or C++ 2011? I'm not sure what to suggest, but it does seem like we should think this through to find a coherent solution before making tweaks we regret.

I have some reservations about defining the __cplusplus macro to 199711L. The fact is, Open Watcom does *not* support C++ 98, at least not fully. So, making that asserting is somewhat a lie. On the other hand, the compiler is at least reasonably close to C++ 98, especially when compared to the other standards.

Right now, there are some C++ 2011 library facilities that are "always on" in the header files regardless of compiler options. I'm thinking that is not a great thing. It might be better to only expose C++ 2011 declarations when the -za0x (or whatever) option is used. In other words, create a fairly strict separation of standards in the library, and in the compiler too. I realize the new stuff could be regarded as an extension on top of the older standard. But for programmers who are less aware of what is in which standard, that mixing seems confusing and undesirable.

So, what I suggest is:

1. An option that activates C++ 98 (with the usual extensions like `near`, `far`, and so on) and disables support for all C++ 2011 features in the language and library. This would be the default mode.
2.  An option that activates whatever meager C++ 2011 support currently exists in both the language and the library, with an obvious and consistent path for adding later standards at some far-flung future time.

jmalak:
> It makes sense to me to change -za0x to something else. That name was chosen in the long-ago days before it was known when C++ 2011 would be finished. The standard was called "C++ 0x" for a long time during its development. Obviously, that is all ancient history.
> 
> An obvious change would be to rename the option za11. The problem, though, is that such a name is ambiguous. Are we talking about C 2011 or C++ 2011? I'm not sure what to suggest, but it does seem like we should think this through to find a coherent solution before making tweaks we regret.
> 
> I have some reservations about defining the __cplusplus macro to 199711L. The fact is, Open Watcom does _not_ support C++ 98, at least not fully. So, making that asserting is somewhat a lie. On the other hand, the compiler is at least reasonably close to C++ 98, especially when compared to the other standards.
> 
> Right now, there are some C++ 2011 library facilities that are "always on" in the header files regardless of compiler options. I'm thinking that is not a great thing. It might be better to only expose C++ 2011 declarations when the -za0x (or whatever) option is used. In other words, create a fairly strict separation of standards in the library, and in the compiler too. I realize the new stuff could be regarded as an extension on top of the older standard. But for programmers who are less aware of what is in which standard, that mixing seems confusing and undesirable.
> 
> So, what I suggest is:
> 
> 1. An option that activates C++ 98 (with the usual extensions like `near`, `far`, and so on) and disables support for all C++ 2011 features in the language and library. This would be the default mode.
> 2. An option that activates whatever meager C++ 2011 support currently exists in both the language and the library, with an obvious and consistent path for adding later standards at some far-flung future time.

It is one of problem to properly handle difference between C and C++ compiler options.
The "similar" problem is with warning messages handling because the numbering is different for C and C++ compiler, but overlap each other. May be we could use prefix "c" for C compiler and "x" for C++ compiler for such type of options.

winspool:
> It makes sense to me to change -za0x to something else. 

c++0x was used by other compiler during C++11 development
c++0x is still supported by autotools

I would leave -za0x as is and probably add "-za11"

> That name was chosen in the long-ago days before it was known when C++ 2011 would be finished. 

And since C++11 development is not finished, it fits perfectly

> The standard was called "C++ 0x" for a long time during its development. Obviously, that is all ancient history.

Yes, a looooong time ago

> An obvious change would be to rename the option za11. The problem, though, is that such a name is ambiguous. Are we talking about C 2011 or C++ 2011? I'm not sure what to suggest, but it does seem like we should think this through to find a coherent solution before making tweaks we regret.

-za11 for wcc*: C11
-za11 for wpp*: C++11

But how does wcl* work?
When detecting file type by extension is enough, 
*.c and *.h for C11 and the others for C++11. 

For owcc, the language is included in "-std=",
but using "owcc -std=c99" fails, when a c++ file should be compiled
(file 99 not found :-)
Can happen in autoconf with "CFLAGS=-std=c99".

> I have some reservations about defining the __cplusplus macro to 199711L. The fact is, Open Watcom does _not_ support C++ 98, at least not fully. 

What about:
-za, -ze or nothing: 1
-za98: 199711L
-za0x: TBD
-za11: 201103L

Then checking, what features are missing for C++98
and switch to 199711L in the feature.

> So, making that asserting is somewhat a lie. On the other hand, the compiler is at least reasonably close to C++ 98, especially when compared to the other standards.

The bug in gcc to change __cplusplus from 1 to 199711L needed 6 years to get fixed
(fix delayed by broken system header on Solaris 8)
and the delay produced many problems on all other systems. 
Fix was changing __cplusplus to 199711L and 
redefine it to 1 in a target specific gcc header for solaris 8.

> Right now, there are some C++ 2011 library facilities that are "always on" in the header files regardless of compiler options. I'm thinking that is not a great thing.

gcc and clang enable everything, what does not conflict with earlier standards, 
produces a warning for features used from newer standards, when "-std=*" is used
and with an additional "-pedantic", the warning is increased to an error.

> It might be better to only expose C++ 2011 declarations when the -za0x (or whatever) option is used. In other words, create a fairly strict separation of standards in the library, and in the compiler too.
> I realize the new stuff could be regarded as an extension on top of the older standard.

Newer features can be used to easier implement features from older standards.
This will not be possible with a hard cut.

> But for programmers who are less aware of what is in which standard, that mixing seems confusing and undesirable.

Only, when using OpenWatcom, as other used compiler behave differently.

> So, what I suggest is:
> 
>     1. An option that activates C++ 98 (with the usual extensions like `near`, `far`, and so on) and disables support for all C++ 2011 features in the language and library. This would be the default mode.

gcc/clang enable extensions by default, or when the value for std starts with "gnu++",
but disables extensions when the value for std starts with "c++" 
(similar for C). 
Old option "-ansi" was relevant before c99: c89 without extensions
That is really simple


For OW, behavior it not always clear.
Example options for C:
none:	boring c94, but are extensions available?
-ze: 	extensions available, but which C version?
-za: 	c94 and extensions disabled
-za99: 	c99, but are extensions available?
-ze -za99  c99, but are extensions still available?
-za99 -ze  extensions available, but still c99 or fallback to c94?
For C++, more combinations are possible.
-zam: unknown, if additional to "-za" or included in "-za"
-zat: unknown, if additional to "-za" or included in "-za"


I suggest this solution, compatible to the old option style:
For C:
-ze: Extensions available, best std possible (or whatever the old style is doing)
-za: extensions disabled, oldest std (or whatever the old style is doing)
-za99: c99, extensions whatever the old style is doing
-za=c99: c99, extensions disabled (new style)
-za=ow99: c99, extensions enabled (new style)
Later C:
-za=c11: c11 mode, extensions disabled
-za=ow11: c11 mode, extensions enabled

For C++:
-ze: Extensions available, best std possible (or whatever the old style is doing)
-za: extensions disabled, oldest std (or whatever the old style is doing)
-zam: use, whatever the old style is doing
-zat: use, whatever the old style is doing
-za0x: c++11 development features enabled, extensions whatever the old style is doing
-za=c++98: c++98, extensions disabled (new style) (implies -zam & -zat)
-za=ow++98: c++98, extensions enabled (new style)
-za=c++0x: c++11 development features enabled, extensions disabled (implies -zam & -zat)
-za=c++0x: c++11 development features enabled, extensions enabled
Later C++:
-za=c++11: c++11 mode, extensions disabled (implies -zam & -zat)
-za=ow++11: c++11 mode, extensions enabled


When "-za=99" is not documented, 
we can separate old option version (std straight after "-za")
and new option version (equal sigh and prefix required after "-za").
Very easy and clear, but only possible, 
if @jmalak is less strict about commandline args.


>     2. An option that activates whatever meager C++ 2011 support currently exists in both the language and the library, with an obvious and consistent path for adding later standards at some far-flung future time.

Works with my suggested solution

winspool:
> @winspool Please create new PR for run-time help fix with correct title.

ok

winspool:
> -za and -ze option has different meaning, it is about C/C++ OW extensions
> There are other -za... options related to some OW features
> Only -za99 for C compiler must be backward compatible (it can be as redundant option)
> That we can design options for C/C++ standards any way
> If we want to use standard version as value then may be suitable something like for C standard -zastdc=89, 99, ... and for C++ standard -zastdx=98, ...

When new commandline options are designed,
please use a style, what is already known to a million user of other compiler,
and what i already suggested (-prefix=value):
-za=c99, -za=c23 / -za=c++98, -za=c++0x, -za=c++11

Please do not reinvent the wheel and make is as complicated as possible....
Sure, -za99 has to stay, but for new options, using different values
by splitting language and year_of_the_sandard is stupid.

Even msvc tries to be similar (/prefix:value) to gcc/clang (-prefix=value),
but the values are the same.
Yes, starting an option with a slash instead of a minus is common on DOS/Windows
and replacing the equal-sign with an column is old DOS style
( "command.com /e:1024"  :-)

msvc:  /std:c++17  /std:c++14  /std:c17  /std:c11
gcc:   -std=c++17  -std=c++14  -std=c17  -std=c11
clang: -std=c++17  -std=c++14  -std=c17  -std=c11

Please do not invent new values.

jmalak:
As I wrote the use of -za and -ze is not possible due to different purpose.
It cannot be change because it break backward compatibility.
The possible is to introduce some new -za... option for it, something like -zastd and setup it's value to looks like -zastd=c99 etc.
Values can be any format it does not matter.

PR #1103, opened 2023-06-21
URL: Replace dead url to help users to find the recent landing Page.

winspool:
(Use https://open-watcom.github.io/ instead of the dead URL)

The old Website looks dead since several Years.

A landing page was restored in Jan.2023 with a download links for the old release of OW 1.9, but nothing more happens since then.

There are more locations in the git sources to the dead URL, (windows resource files, example sources, documentation and more), but that might need discussion first.

--
Regards ... Detlef

### DIFF
--- a/bld/watcom/h/banner.h
+++ b/bld/watcom/h/banner.h
@@ -65,7 +65,7 @@
 #define banner2a(year)      banner21a(year) " All Rights Reserved."
 
 #define banner3             "Source code is available under the Sybase Open Watcom Public License."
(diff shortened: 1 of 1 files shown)


jmalak:
thank you for your contribution.
I rework your PR to include also documentation.
the binaries use reference to OW v2 Github readme page
documentation use 3 URLs which are configured in docs/mif/master.mif make file
**`owurl`** variable point to OW v2 Github readme page
**`owurl1`** variable point to OW v2 Github issues page
**`owurl2`** variable point to OW v2 Discord Channel page

PR #1108, opened 2023-06-28
Add Windows NT for MIPS support in clib

lakor64:
The following PR adds Windows NT for MIPS in WATCOM C library, the compilation doesn't currently fully work due to missing CG code (which will be pushed in separate PRs)

- Added clib projects for Windows NT mips
- Added generation of export.tmp (-z flag) for RISC platforms (mps, ppc, axp)
- Implement NT MIPS startup (same as AXP for now)
- Fix heap compilation on non-8086 platform
- Add missing flags to use NT api on MIPS inside libwin32
- Add TGT_SYS check on mthread
- Stub fpu exception implementation for now (I will investigate them and make a PR to either specifiy there's no fpu exceptions or implement the relative code, or in this PR in case it's required)

PR #1109, opened 2023-06-28
Build mathlib for Windows NT MIPS

lakor64:
Similar style of #1108 , this PR adds the Windows NT for MIPS platform for mathlib, it also adds a missing check for MIPS on watcom project (which fixes crt build)


jmalak:
Thank you for your contributions.

PR #1112, opened 2023-07-07
Updates rtdll for RISC platforms

lakor64:
The following commits fixes rtdll compilation for NTMIPS (currently the linker is broken at reading .lib file, under investigation)

I have also added preliminary support for ntppc and rtdll for ntaxp, just in case someone wants to continue working for those platforms.


jmalak:

PR #1114, opened 2023-07-09
Add mips checks for mstools

lakor64:
The following commit adds MIPS platform checks for mstools converter


jmalak:

PR #1117, opened 2023-07-10
Portability fixes for bootstrap OW on ppc-linux

winspool:
This patch allows starting to bootstrap OW on ppc-linux

(bootstrap halts in "bld/causeway/cwc/dosi86", 
when OW try to use "wclpcc" (release version)) 

-- 
Regards ... Detlef


bld/plusplus/h/pcheader.h
jmalak, line 1: The correct architecture macro is __PPC__.
don't use redundant architecture macros.
I added Wiki page [Source Code Architecture Macros](https://github.com/open-watcom/open-watcom-v2/wiki/Source-Code-Architecture-Macros)
please fix it.

Issue #66, opened 2014-07-18
switch...case and 64-bit integer types


ArmstrongJ:
I'm running into some code that's failing because the argument to a switch statement is of the type _uint64_t_ (via a typedef).  I'm assuming such code is not allowed under c89 (?).  However, I'm really not sure.

Not sure this would be a bug, possibly more of an enhancement.  It would be nice if OpenWatcom's switch...case supported any integral value.

sezero:
This issue is standing in the way of building current nasm with watcom.

vweisner:
XZ 5.4.1 fails to build because of this issue. Are there any new updates on fixing this issue?

hpax:
> This issue is standing in the way of building current nasm with watcom.

NASM maintainer here.

Right now we are carrying some ugly hacks for OW to hack around this, but at some point soon we are going to have to drop support for compilers which don't support all of C99, and this is *by far* the worst painpoint.

rdos314:
It's a big issue with OpenSsl as well, where people are adding code that uses 64-bit integers, and which needs to be selectively patched. I doubt they will change this since other compilers don't have an issue with it, and they don't even notice it since they don't build their software with OW.

Also note that all of this code have case values less than 32-bits, so even a fix that just truncates the long long value to int would be much better than the current compile problem.

jmalak:
I started work on Code generator changes that now it is nearly finished.
I need to design some code generator run-time support for 64-bit values scan and jump tables and also add some code generator checks to not overflow 16-bit or 32-bit architecture limits.
I did a few changes in C compiler front-end to remove some residual code which still use 32-bit values instead of 64-bit even if for 32-values.
code for front-end constants handling in C compiler need some rework to be completely 64-bit and doesn't mix 32-bit and 64-bit processing.
C++ compiler should be enhanced by some C99 features as extension (mainly constants processing) and next fix for "64-bit switch"
Fortran compiler will be relatively simple, only enhance SELECT and computed GOTO statements to support 64-bit types as argument.

jmalak:
#1455 same issue

jmalak:
From now OW should support 64-bit in switch statement (for expression or case value). 
Please recheck your issue with latest OW version.

PR #1118, opened 2023-07-12
unsigned long long support in switch statements (cstmt.c)

vweisner:
This commit is designed to fix a C compiler limitation on switch statements. In the cstmt.c file, I added support for unsigned long long to fix issue #66.

### DIFF
--- a/bld/cc/c/cstmt.c
+++ b/bld/cc/c/cstmt.c
@@ -1002,6 +1002,10 @@ static void SwitchStmt( void )
         sw->case_format = "%lu";
 //        switch_type = TYP_ULONG;
         break;
+    case TYP_ULONG64:
+        sw->case_format = "%llu";
+//        switch_type = TYP_ULONG64;
+        break;
     case TYP_LONG:
 //        switch_type = TYP_LONG;
         break;
--- a/bld/cg/c/bestub.c
+++ b/bld/cg/c/bestub.c
@@ -106,8 +106,8 @@ extern void CGControl(cg_op _1,cg_name _2,label_handle _4) {}
 extern void CGBigLabel(back_handle _4) {}
 extern void CGBigGoto(label_handle _1,int _4) {}
 extern sel_handle CGSelInit(){return(0);}
-extern void CGSelCase(sel_handle _1,label_handle _2,signed_32 _4) {}
-extern void CGSelRange(sel_handle _1,signed_32 _2,signed_32 _3,label_handle _4) {}
+extern void CGSelCase(sel_handle _1,label_handle _2,signed_64 _4) {}
+extern void CGSelRange(sel_handle _1,signed_64 _2,signed_64 _3,label_handle _4) {}
 extern void CGSelOther(sel_handle _1,label_handle _4) {}
 extern void CGSelectRestricted(sel_handle _1,cg_name _4,cg_switch_type _2 );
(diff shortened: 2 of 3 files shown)


jmalak:
I am sorry, but it is not a fix.
It only mask a problem, by suppressing appropriate message in C compiler.
The reason why 64-bit value can not be used by any OW Compiler Front-end is that OW Code Generator doesn't support to generate such code which handle 64-bit value in switch statement.
It requires to fix Code Generator to be able generate appropriate code for 64-bit value it includes appropriate optimization etc. Look into [OW Code Generator documentation](https://open-watcom.github.io/open-watcom-v2-wikidocs/cgdoc.html) for section "Select and Switch statements". CG API function to generate appropriate code start with CGSel... There is also example how to use it. All these functions handle only 32-bit values, they must be extended to 64-bit values or add new one for 64-bit values. After that you can modify C compiler Front-end to use these new/modified API function for 64-bit value switch construct. The same must be done for C++ compiler Front-end.

vweisner:
I have redefined the 32-bit types to 64-bit for the code generator, however it seems that the C++ compiler is more complex than the C compiler when it comes to the switch statement code. Is there a function or file that I can be directed to that would need to be modified in order for the C++ compiler to operate the switch statement correctly?

The documentation for the code generator has not provided additional information about the functions, except the parameter types of each function. I will look at the example to see if there is anything else I need to add to the code generator.

jmalak:
I am sorry you are naive if you think that only API change fix the issue.
If code is compiled correctly then it doesn't mean that all is working correctly.
What I talk about CGSel... functions it means to rework them to realy handle 64-bit value.

!!!! First you need understand what you are doing and understand appropriate part of Code generator !!!!!

Code Generator is complex code with multiple processing layers which are responsible for approriate tasks.
You need study appropriate parts of CG code and read appropriate pages in Wiki and documentation.
The extension of switch processing to 64-bit is not simple task.
I have it on my to-do list for long time because I still not understand all piece of code which is involved in switch processing.
You should take attention on `BGSel...` functions which are processing functions, called from `CG API CGSel...` functions.
You need to understand at minimum optimization for switch constructs, look for usage of enum items `CG_SWITCH_SCAN`, `CG_SWITCH_BSEARCH` and `CG_SWITCH_TABLE`.
Most important is BGSelect function which process full switch statement, all other functions are only to define cases and init/fini switch processing.
It is clear that CG_SWITCH_SCAN optimization is not possible for 64-bit values because it is capable handle only 32-bit value.
Use of this optimization should probably be suppressed or rework for 64-bit values, but not for narrower integer types.
Take into acount that code must work for both 16-bit and 32-bit target architecture. There are different limits for 16-bit and 32-bit architectures.

Please don't submit such WIP work.
The change to support 64-bit value in switch statement is complex change into crucial part of OW (Code Generator) that without intensive testing and without full pass of all OW tests will not be accepted.

vweisner:
I have very limited time on what I can do. It may be a WIP project, but that does not mean others can't help when help is needed. I am not naive, as you said because my limited time and knowledge of the project is very minimal. To be quite frank, I did not think there would be so much to do. I will work on the project as much as I can to get this done. I will close the pull request now, and work on it, then open a new one when I finish the code.

jmalak:
You can open issue and consult your question/code/change in this issue.
If you have your repository on Github that it is possible to contribute to your branch. 
I can help you with anything what I know or contribute to your code changes.

PR #1119, opened 2023-07-13
ide: Fix the Dialog type, when saving the build logfile

winspool:
I was surprised, that the action Button had the label "Öffnen" (Open), after selecting "Save As.." to save the Build Logfile.

Tested on Linux with Wine

(The WFileDialog class is implemented in bld/wclass/cpp/wfiledlg.cpp)

--
Regards ... Detlef


winspool:
Before my fix:

winspool:
Before my fix:

jmalak:
It is very old typo and nobody report it.

PR #1120, opened 2023-07-17
linuxppc: Declare a linuxppc target for wlink and owcc

winspool:
That's far from working, but a mini step.

No arch specific headers or runtime libraries for ppc-linux are available.

--
Regards ... Detlef


jmalak:

PR #1132, opened 2023-09-11
WIP: C++ compiler - changes for ppc and mips architectures

winspool:
I used the same size as ow is using in the c compiler

--
Regards ... Detlef

### DIFF
--- a/bld/plusplus/c/cgbkmain.c
+++ b/bld/plusplus/c/cgbkmain.c
@@ -220,6 +220,10 @@ static TRY_IMPL* tryImpl(       // FIND/ALLOCATE A TRY IMPLEMENTATION
     #define JMPBUF_SIZE ( 24 * TARGET_DOUBLE )
 #elif _INTEL_CPU
     #define JMPBUF_SIZE ( 13 * TARGET_INT )
+#elif _CPU == _MIPS
+    #define JMPBUF_SIZE ( 26 * TARGET_LONG )
+#elif _CPU == _PPC
+    #define JMPBUF_SIZE ( 1 * TARGET_INT )
 #else
     #error bad target
 #endif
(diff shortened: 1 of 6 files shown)


winspool:
For the rest, i mostly changed "_CPU == _AXP" to "_RISC_CPU",
but is that the prefered way to solve the rest of the build failures?

* Probably defining and using a "_32BIT_CPU" is better

* With my change to "#if _INTEL_CPU || ( _RISC_CPU )",
  the related code path is always used.
  Is removing the "#if" check a better solution? 


By applying the rest of the patchset, wppmps and wppppc
are build and start to work.
 
[0000_rest_for_wppmps_and_wppppc.txt](https://github.com/open-watcom/open-watcom-v2/files/12580354/0000_rest_for_wppmps_and_wppppc.txt)

jmalak:
We don't want to mask the problems, the problems must be solved.
Replacing the Alpha architecture condition with the _RISC_CPU macro is not a good idea.
Each such condition has a reason, you need to study the code and decide on a solution.
AXP, PPC and MIPS are different architectures with different features.

winspool:
> We don't want to mask the problems, the problems must be solved.
> Replacing the Alpha architecture condition with the _RISC_CPU macro is not a good idea.

That was also my idea, and for that reason, the problematic "_RISC_CPU" changes
are not included in this pull request,
but i attached them as a diff, so everyone can see the current status.

Please commit the clean changes in this pull request 
to make research and testing a lot easier

> Each such condition has a reason, you need to study the code and decide on a solution.
> AXP, PPC and MIPS are different architectures with different features.

I will do that.

jmalak:
Please rebase it to OW2 HEAD and fix a problems if any.

jmalak (commented):
If any change related to precompiled header file processing then change minor version to invalidate already created PCH

PR #1136, opened 2023-09-16
owcc: Accept pthread and pedantic as alias for mthreads and Wextra

winspool:
Most unix based Open Source Projects expect gcc or clang as compiler.
(or the gcc port mingw on windows)

OpenWatcom is old and the userbase is so small compared to gcc or clang,
that porting the affected projects to support OpenWatcom is nearly impossible.

Sometimes the only Problem when using owcc instead of gcc/clang
is a commandline option, which is not handled by owcc.

By making owcc commandline handling less strict,
OpenWatcom can be used more often as a dropin for gcc or clang.

This patch allows owcc to accept "-pthread" and "-pedantic":

-pthread option in gcc / clang:
* defines _REENTRANT (obsolete, was selecting multithread aware implementations)
* links with libpthread

With this patch, "-pthread" is an alias for the windows based option "-mthreads"
For linking, no adjustment is needed, as the
OpenWatcom runtime library already includes the pthread code.

-pedantic option in gcc / clang
* Use more checks as with "-Wall" and "-Wextra"

With this patch, "-pedantic" is an alias for "-Wextra".

--
Regards ... Detlef

### DIFF
--- a/bld/wcl/c/owcc.c
+++ b/bld/wcl/c/owcc.c
@@ -235,6 +235,7 @@ static option_mapping mappings[] = {
     { "mwindows",                       "bg" },
     { "mconsole",                       "bc" },
     { "mthreads",                       "bm" },
+    { "pthread",                        "bm" },
     { "mrtdll",                         "br" },
     { "mdefault-windowing",             "bw" },
     { "mhard-emu-float",                "fpi" },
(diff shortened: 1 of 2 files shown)


winspool:
The commandline help for owcc is alphabetic sorted in a group.

When I add `pedantic` for the commandline help above or below `Wextra`,
the sorting moves `-pedantic` between `-mwindows` and `-shared`.

```

:option. pedantic
:usage. enable maximal warnings
:group. 3
:target. any

```

Is there a way to avoid that sorting?

jmalak:
We discused it several time and we will not create another "bad" clone of GCC/CLANG.
We are orinted on ISO C standard compliance and POSIX where it has sense.
!!!! owcc is not GCC/CLANG driver it is POSIX CC driver. !!!!!
If you want compatibility with GCC then you can create similar utility as owcc but with compatibility with GNU GCC.

If you wan't compile GNU projects then you should use GNU tools otherwise you must port it to OW.
It is similar as request to Microsoft to compile GNU projecs.

PR #1152, opened 2023-11-15
Fixing D3DTEXTUREMAGFILTER member names.

americusmaximus:
Fixing D3DTEXTUREMAGFILTER member names:

1. D3DTFG_GUASSIANCUBIC -> D3DTFG_GAUSSIANCUBIC
2. D3DTFG_ANSIOTROPIC -> D3DTFG_ANISOTROPIC

to match original names in DirectX SDKs.

Also removing check for 0x600 version. There is no such version check in DirectX 6.0 SDK, and those values definitely were supported in at least DirectX 5.0.

### DIFF
--- a/bld/w32api/include/directx/d3dtypes.mh
+++ b/bld/w32api/include/directx/d3dtypes.mh
@@ -1485,6 +1485,7 @@ typedef enum _D3DTEXTUREOP {
 #endif
 
 /* Direct3D texture magnification filters */
+#if (DIRECT3D_VERSION >= 0x0600)
 typedef enum _D3DTEXTUREMAGFILTER {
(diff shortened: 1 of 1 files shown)


jmalak (commented):
I checked latest MS d3dtypes.h and there is "#if (DIRECT3D_VERSION >= 0x0600)" preprocessor guard. I think we should hold it here.
I checked DirectX 5 header files and they don't contain such definitions. Look like it was introduced by DirectX 6 without guard which was probably added later.

americusmaximus:
Yes, you are correct, my mistake. Brought the guard back. Thanks.

jmalak:
Thanks for your contribution.

PR #1174, opened 2023-12-23
Windows header WINVER #ifdefs based on Windows 1.x and 2.x SDK headers

joncampbell123:
Allow WINVER=0x300 for mmsystem.h because Windows 3.0 did add the MMSYSTEM API after release and a symbol dump of MMSYSTEM.DLL shows most of the API is there in Windows 3.0 Multimedia Edition


jmalak:
Many Thanks for your contribution.

PR #1175, opened 2023-12-24
devguide: Fix binbuild path, Update arch for as and cg

winspool:
The default subdirectory created during the "boot" and the "bootstrap" phase
and used later was changed to "binbuild" a while ago.

Add infos about additional arch to the "as" and "cg" projects

--
Regards ... Detlef


jmalak:

PR #1177, opened 2023-12-24
Some MF_ constants like MF_OWNERDRAW did not appear until Windows 3.0

joncampbell123:
Some resource flags including DS_SETFONT did not appear until Windows 3.0. Include a note stating that DS_SETFONT will cause Windows 2.x to fail to load the dialog box and why


jmalak:

PR #1183, opened 2023-12-25
Windows 2x submenu check icon translation wrc 20231225 0142

joncampbell123:
This change allows existing ICO resources to work with Windows 2.0 provided that one of the formats contained in the ICON directory is a 64x64 monochrome image.

This also adds a warning if submenus are listed in any MENU resource when targeting Windows 2.0.

### DIFF
--- a/bld/rc/rc/c/semmenuw.c
+++ b/bld/rc/rc/c/semmenuw.c
@@ -271,6 +271,21 @@ static bool SemWriteSubMenu( FullMenu *submenu, int *err_code, YYTOKENTYPE token
     return( error );
 }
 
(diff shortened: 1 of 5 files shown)


joncampbell123:
Microsoft's Win16 tools for Windows 3.1 will not make a 64x64 monochrome icon (doesn't give you the option), but many open source tools including ImageMagick and the GIMP are happy to let you make such ICO files.

To combine multiple ICO files together into one (such as the normal 32x32x4bpp and the 64x64x1bpp needed for Windows 2.0) you may use a Python script I whipped up to join them together.

https://github.com/joncampbell123/doslib/blob/master/tool/icojoin.py

PR #1185, opened 2023-12-25
Resource compiler: CURSOR resource translation for Windows 2.x targets

joncampbell123:
This code also splits the image+mask code out into a common function for both CURSOR and ICON translation.


joncampbell123:
Build failed because of bld/ssl? This commit never changes that. What?

joncampbell123:
Later tonight perhaps around midnight my time (PST) I intend to submit another pull request to add BITMAP resource translation to the resource compiler. It should be just like ICO/CUR translation but without the mask.

PR #1190, opened 2023-12-29
DirectX headers list each C-level method call twice.

joncampbell123:
Let's use variadic macros to try to clean that up and improve maintainability


jmalak:

PR #1191, opened 2023-12-29
trmem: Fix header width on 64bit Linux builds

winspool:
The header for a memory leak report from bld/trmem 
is better to read  with the correct wide.

Example from building hcwin/nt386:
```
bwcc386 -zq -D_BLDVER=1300 -D_CYEAR=2023  -j-we -od -d2 -x-wpx -xx -zam -fo=trmem.obj -fpi -mf -zc -bt=nt -DWINVER=0x400 -D_WIN32_IE=0x300       -wx-wce=C310-zastd=c99 -D_ENABLE_AUTODEPEND         -I"../h"  -I"../../../bld/hdr/dos/h" -I"../../../bld/w32api/nt/h" -I"../../../bld/trmem" -I"../../../bld/watcom/h" ../../../bld/trmem/trmem.c
%write whc.lnk debug dwarf all op nored op symfile op map sys nt
bwlink op q name whc.exe @whc.lnk
Current usage: 0000000000000314 bytes; Peak usage: 0000000000226140 bytes
  Who              Addr             Size     Call             Contents
================ ================ ======== ================ ===========================
0000000000000000 00005625daccecc0 00000019 0000000000001eec 5f5f776370705f34 5f756e646566__wcpp_4_undef
0000000000000000 00005625dad5be80 00000019 0000000000001eeb 5f5f776370705f34 5f756e646566__wcpp_4_undef

```

--
Regards ... Detlef

### DIFF
--- a/bld/trmem/trmem.c
+++ b/bld/trmem/trmem.c
@@ -92,10 +92,16 @@ msg(MIN_ALLOC,          "%W allocation of %T less than minimum size" );
 #elif defined( _M_I86LM ) || defined( _M_I86HM )
     msg(PRT_LIST_1,     "   Who      Addr    Size   Call   Contents" );
     msg(PRT_LIST_2,     "========= ========= ==== ======== ===========================================" );
+#elif defined( __SIZEOF_POINTER__ ) && ( __SIZEOF_POINTER__ == 8) && (__SIZEOF_LONG__ == 8 )
+/* Building OW on a 64bit Host with gcc/clang and similar compiler: a long is 64bit width */
+    msg(PRT_LIST_1,     "  Who              Addr             Size     Call             Contents" );
+    msg(PRT_LIST_2,     "================ ================ ======== ================ ===========================" );
 #elif defined( _WIN64 )
+/* Building OW on Win64: a long is only 32bit width */
     msg(PRT_LIST_1,     "  Who              Addr             Size     Call     Contents" );
     msg(PRT_LIST_2,     "================ ================ ======== ======== ===========================" );
 #else
+/* Default: Building OW on a 32bit Host */
     msg(PRT_LIST_1,     "  Who      Addr     Size     Call     Contents" );
     msg(PRT_LIST_2,     "======== ======== ======== ======== ===========================================" );
 #endif


jmalak:
you introduce new macros which are "foreign" for OW build system and is not portable.
They are
__SIZEOF_POINTER__
__SIZEOF_LONG__
They are not defined by any standard or by OW build system
it is necessary to change it to be portable

winspool:
> you introduce new macros which are "foreign" for OW build system and is not portable.
> They are
> **`__SIZEOF_POINTER__`**
> **`__SIZEOF_LONG__`**
> They are not defined by any standard or by OW build system
> it is necessary to change it to be portable

Dou you have a suggestion, what to use instead?
A "sizeof(long)" or similar does not work in the preprocessor.


For gcc/clang/tcc and all other compiler compatible to them,
i can use the compiler-driver with the args " -E -dM " 
to dump all predefined macros. That works also with addition args.

(Unfortunately, i found nothing similar for OpenWatcom.)

Examples:
```
echo ""  |  tcc  -E  -dM -  |  sort
tcc  -m32  -E  -dM  -  </dev/null   |  sort

gcc  -march=native  -E  -dM  -  </dev/null   |  sort
X86_64-w64-mingw32-gcc  -E  -dM -  |  sort

```

For the simple tcc, the resulting list is small. (31 entries).

Since many comercial compiler are based on clang,
(Examples: Embarcadeo, Visual Studio, Intel, IBM),
this works also for them.

jmalak:
it is not about names. First you should look into OW code how this situation is solved and follow it, don't introduce new construct if it already exists. Don't try to introduce any GNU or other toolchain macros if they are not part of standard (ISO, POSIX) to OW build system, it must be independent on toolchains, but it can be macros used by OW toolchain. why you don't use existing OS and architecture global macros? you are able determine if platform is 32-bit or 64-bit and what OS is used. by example 64-bit Linux can be determined by existing macro **`__LINUX_X64__`**.

jmalak:
I fixed a one bug and use _M_X64 macro instead of _WIN64
Anyway using it on 64-bit has no sense, it is design for 16/32-bit Watcom platforms only.

Issue #1200, opened 2024-01-03
WCC 80-bit long double support -fld is not supported by the code generator


joncampbell123:
If the -fld switch is used to enable the full 80-bit long double format, the compiler will either crash with an internal error or produce buggy code that causes crashes if any function passes a long double as a parameter to a function call.

Perhaps this issue has been lurking for awhile considering that -fld is not listed by the compiler when printing out all the options, and that the compiler *by default* treats long double as an alias to double (as Microsoft C++ does) meaning both data types are the 64-bit double floating point precision data type by default.

The problem seems to stem from code within bld/cg that is called assuming that everything it is asked to resolve or handle is 8 bytes or less when deciding which registers or stack arrangement to do. 10 bytes are needed for 80-bit long double.

joncampbell123:
More details while discussing this request: https://github.com/open-watcom/open-watcom-v2/pull/1198

Test case: https://github.com/joncampbell123/doslib/blob/master/hw/dos/testprna.c

joncampbell123:
The design of the x86 CG seems to indicate that it was never intended to handle the FL type, and there is a lot written around handling "long double" as 64-bit double. This will require a lot of steady hacking.

Also I noticed that sizeof(long double) == 16 when -fld is given. 6 bytes of padding.

joncampbell123:
Latest test: The same internal error occurs if I try to pass the address of a long double constant to a function call, i.e. declare a long double constant "a" (which doesn't trigger the error) and  then pass &a to a function with a parameter of type "long double *".

joncampbell123:
I can't even declare the constant and try to pass it off as typecast (unsigned char*)(&lv) :laughing: oh come on now!

   static void TEST(const unsigned char *x,size_t sz) {
      ...
   }

   #define TEST(x) {\
        const double dv = x;\
        const long double lv = x ## l;\
        TEST((const unsigned char*)(&dv),sizeof(dv)); \
        TEST((const unsigned char*)(&lv),sizeof(lv)); }

joncampbell123:
Putting the two into a struct makes it worse because taking the address of the struct triggers the error too.

I finally found the one way that OW allows me with -fld to declare a long double and obtain it's memory address though.

Printing out the bytes in memory, it looks like sizeof(long double) == 10 as expected but OW is really just formatting it as 64-bit double with an extra 16 bits of zero.

https://github.com/joncampbell123/doslib/blob/master/hw/dos/testprnb.c

[...]

PR #1205, opened 2024-01-05
__declspec(fld) to enable 80-bit long double (same as if -fld were given) but on a symbol by symbol basis

joncampbell123:
Allow compiled C/C++ code to enable 80-bit long double on a symbol by symbol basis by adding __declspec(fld) to the data type. Allow the __declspec(fld) to function the same as if -fld were enabled at that very point.

This can help enable support for 80-bit long double gradually through the codebase without breaking anything.

Related to https://github.com/open-watcom/open-watcom-v2/issues/1200

### DIFF
--- a/bld/cc/c/ctype.c
+++ b/bld/cc/c/ctype.c
@@ -311,7 +311,7 @@ static TYPEPTR GetScalarType( bool *plain_int, type_mask bmask, type_modifiers f
         data_type = TYP_FLOAT;
         break;
     case M_LONG | M_DOUBLE:
-        if( CompFlags.use_long_double ) {
+        if( CompFlags.use_long_double || (flags & FLAG_FLD) ) {
             data_type = TYP_LONG_DOUBLE;
             break;
         }
@@ -323,7 +323,7 @@ static TYPEPTR GetScalarType( bool *plain_int, type_mask bmask, type_modifiers f
         data_type = TYP_FCOMPLEX;
         break;
     case M_COMPLEX | M_LONG | M_DOUBLE:
-        if( CompFlags.use_long_double ) {
+        if( CompFlags.use_long_double || (flags & FLAG_FLD) ) {
             data_type = TYP_LDCOMPLEX;
             break;
         }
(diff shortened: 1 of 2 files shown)


joncampbell123:
See also this DOSLIB test program which was updated to test it:

https://github.com/joncampbell123/doslib/blob/master/hw/dos/testprnb.c

db1

joncampbell123:
If "fld" is not a good __declspec name, or there's a better way to do it, let me know.
The idea is to allow the use of 80-bit long double even if -fld isn't specified by prefixing "long double" with something that overrides the default long double == double mapping. Internal parts of OW could use this to properly handle long double without needing the -fld flag.

jmalak:
The purpose of the -fld option is to allow a long double to be used as an 80-bit FP including all math functions, etc.
By default long double type is mapped to double type (by FE) all in-line FP math (80bit) is rounded to 64bit FP.
With the -fld option, long double type is not double, but an 80-bit FP type, and all in-line FP math is not rounded to just 64-bit FP and stay with full 80-bit precission.
CRTL gets information from the compiler about what size long double to use and switches all math functions between 64-bit or 80-bit for the long double type and any appropriate math functions.
That something like __declspec(fld) cannot be implemented.

joncampbell123:
> The purpose of the -fld option is to allow a long double to be used as an 80-bit FP including all math functions, etc.
> By default long double type is mapped to double type (by FE) all in-line FP math (80bit) is rounded to 64bit FP.
> With the -fld option, long double type is not double, but an 80-bit FP type, and all in-line FP math is not rounded to just 64-bit FP and stay with full 80-bit precission.
> CRTL gets information from the compiler about what size long double to use and switches all math functions between 64-bit or 80-bit for the long double type and any appropriate math functions.
> That something like __declspec(fld) cannot be implemented.

The expectation here is only to be able to declare such a variable, even if at this time you can do little else with it due to total lack of support in the CG. I don't know how you expect to fix it without such a feature then.

If use of the declspec triggered a one time warning about lack of math runtime support, would you allow this?

PR #1212, opened 2024-01-25
c_guide: Add Linux and more Windows versions

winspool:
The old suggestion to provide telephone and fax number for service was also replaced by a suggestion to provide a valid email address.

I expect, that today no one will send a fax or make a phone call
to help fixing a bug.

Is this ok this way, or should the hint "experimental" be added in some locations?


winspool:
I did not rebuild the documentation yet and i hope,
that my minimal changes will not break anything.

PR #1236, opened 2024-02-16
unistd.h: Support also _SC_PAGE_SIZE for sysconf()

winspool:
The value is the same as _SC_PAGESIZE (as allowed by POSIX)
to reduce the implementation codesize to 0 Byte.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/unistd.h.html

_SC_PAGE_SIZE is required by POSIX and needed to compile zycore-c

--
Regards ... Detlef

### DIFF
--- a/bld/hdr/watcom/unistd.mh
+++ b/bld/hdr/watcom/unistd.mh
@@ -62,7 +62,7 @@
 #endif
 
 /* Symbolic constants for sysconf */
-/* caution: the module sysconf.c bound checks _SC_ARG_MAX ... _SC_PAGESIZE */
+/* caution: the module sysconf.c bound checks _SC_ARG_MAX ... _SC_THREAD_STACK_MIN */
 #define _SC_ARG_MAX                         1
 #define _SC_CHILD_MAX                       2
 #define _SC_CLK_TCK                         3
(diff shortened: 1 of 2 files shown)


bld/hdr/watcom/unistd.mh
jmalak, line 1: It is not your fault, but implementation note should be as comment in source code or in header source as programmer comment (no value for user, rather confusing)
comment in header source start with two double dots "::", it is not output to header
it can looks like
::
:: caution: the module sysconf.c bound checks _SC_ARG_MAX ... _SC_THREAD_STACK_MIN
::

winspool: I did not know, that comments are handled special during conversion from *.mh to *.h.

Just saw the outdated bounds check value near my fix and updated it.

Can you fix it after committing, or should i resend without touching the comment?
(When you change the comment afterwards, i can learn, how it is supposed to work.

Or should I try to update the comment myself and resend the patch?

jmalak: simply change it. You or I can squash your changes in this pull request as single commit.

jmalak:
Thanks for your contribution

PR #1268, opened 2024-04-07
fix wdump when symbol name length > 0x7f

Arcnor:
As the title says, when a symbol name length exceeds `0x7f` the calculation for size and next position becomes negative and everything breaks.


jmalak:

PR #1273, opened 2024-04-19
Fix D3DRGBA() macro

thp:
The definition of the `D3DRGBA()` macro had unbalanced parentheses.

This change balances them out and also improves formatting, so it's easier to see what goes where.


jmalak:

PR #1275, opened 2024-04-22
hdr: Add sys/select.h posix header

winspool:
Detected, when i tried to compile TinyEMU with OpenWatcom.

In early posix versions, these declarations / defines / macros where in `sys/time.h`,
but since some posix revisions, they belong to `sys/select.h`.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_select.h.html

I tested it on linux, but i can`t test RDOS

--
Regards ... Detlef


jmalak:
I have a few notes:
- implementation comment should not be output to header file, comment defined with two double dots "::" on begining of line
- why you write such nonsense to header files, see below, POSIX requires to be defined in both, that OpenWatcom sys/time.h is OK
/* Posix requires, that this belongs to sys/select.h
 * OpenWatcom had it previously in sys/time.h
 */
 what value it has for users or programmers, only increase size of header files
may be better to define fdset.sp file and include it in both

jmalak:
I added fdset.sp and select.sp include files (now used in sys/time.h) holding shared definition that only sys/select.h must be implemented

winspool:
Updated patch on top of the latest changes in the master tree

jmalak (commented):
Please update it against latest sources
time_t.sp is now included in timeval.sp that code can be simplified.
sys/select.h is build only for LINUX and RDOS that it is not necessary guard for them, only LINUX specific thing must be guarded.

PR #1280, opened 2024-04-29
owcc: Disable the banner with the option -zq

winspool:
owcc already recognize the parameter "-zq".

With this patch, "-zq" is used to disable the banner.
("-zq" also used by the OpenWatcom compilers to disable the banner)

More options to disable the welcome banner (not implemented for owcc):
* The OW compiler use "-q" as an alias for "-zq"
* Other Toolchains for Windows support "-nologo" to disable the welcome banner 

--
Regards ... Detlef

### DIFF
--- a/bld/wcl/c/owcc.c
+++ b/bld/wcl/c/owcc.c
@@ -161,6 +161,7 @@ static  bool        cpp_want_lines;         /* flag: want #lines output?        
 static  bool        cpp_keep_comments;      /* flag: keep comments in output?     */
 static  bool        cpp_encrypt_names;      /* flag: encrypt C++ names?           */
 static  char        *cpp_linewrap;          /* line length for cpp output         */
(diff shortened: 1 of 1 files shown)


jmalak:
what sense has this change?
check owcc code for -v and -zq options

winspool:
> what sense has this change?
> check owcc code for -v and -zq options

Without any extra args, there is no output from owcc.

To see the programs, which are called by `owcc`,
"-v" has to be used,
but when "-v" is used, every program also print it's banner.

A solution, which is used by other compiler to disable the banner
is a new arg "-nologo".

With my patch, "-zq -v" can be used
to see every program called by `owcc`,
but avoids every additional output.
(after wlink is patched similar to disable the banner)

Reusing "-zq" is a far smaller changeset
than adding a new argument.

PR #1282, opened 2024-05-02
wdump:  Remove unused defines

winspool:
Found while study the wdump source

-- 
Regards ... Detlef

### DIFF
--- a/bld/exedump/h/wdglb.h
+++ b/bld/exedump/h/wdglb.h
@@ -49,12 +49,10 @@
 #include "lib.h"
 
 #define PERLINE                 16
-#define NUMLINE                 24
 #define LINE_LEN                78
 #define SLEN                    30
 #define STDOUT                  1
 #define BSIZE                   512
-#define MAX_SECTION_DATA        4096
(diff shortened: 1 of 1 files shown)


jmalak:
what's the point?
are you sure what these macros mean and if they won't be used anymore?

winspool:
> what's the point?

While working on other projects many years before,
I was teached to make small, independant patches.

> are you sure what these macros mean

I try to understand better, how the code is working
to extend wdump and found out,
that these defines are unused.

> and if they won't be used anymore?

I canged the replacement value to a non existent function call
and rebuild OpenWatcom without a build break.

> In OW code base is many unused macros, enumeration values.
> do you plan remove all?

I do not search for them,
but when i touch a code area and find someting unused,
i remove them.

jmalak:
It is little bit stupid arguments.
General OW rule is don't tuch what is not part of your change.
That something is not used is not argument to remove it.
Such thing could be source of information about code evolution, design etc.
You cannot know why it is where it is and reason for it.
Don't remove such unrelated definition anytime.

PR #1292, opened 2024-05-20
rc/exedmp: Build the exedmp tool also on i386-linux

winspool:
This tool uses Microsoft specific runtime functions
and needs more fixes for 64bit.

--
Regards ... Detlef

### DIFF
--- a/bld/rc/exedmp/builder.ctl
+++ b/bld/rc/exedmp/builder.ctl
@@ -15,10 +15,11 @@ set PROJDIR=<CWD>
 
 [ BLOCK <BLDRULE> rel cprel ]
 #============================
-    <CCCMD> os2386/exedmp.exe  "<OWRELROOT>/binp/"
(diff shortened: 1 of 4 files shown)


jmalak:
What sense has OW internal resource tools to be build for Linux if resource compiler can be developed only for Windows and OS/2 and can be tested only on these OS-es?

winspool:
I am working on extending wdump to better support 64bit
and try everything available to look inside any file.

jmalak:
It is OK to extend it to PE 64-bit, but we are talking about Linux build of rcdmp which has no sense.
Anyway wdump already support 64-bit PE header, see exedump/c/peexe.c source file.

winspool:
> It is OK to extend it to PE 64-bit, 
> but we are talking about Linux build of rcdmp which has no sense.

Compiling for Windows is most times done on Linux here.

> Anyway wdump already support 64-bit PE header, see exedump/c/peexe.c source file.

64bit elf support is missing
(64bit macho for looking in armo64 and bino64 is a bonus)

PR #1306, opened 2024-07-16
wdump: Fix Elf64 header output

winspool:


jmalak:
thanks for your contribution

PR #1312, opened 2024-07-29
whelp: Avoid a crash, when the search path has an empty segment

winspool:
The double colon in $PATH let whelp segfault,
when whelp is searching for available help files.

--
Regards ... Detlef

### DIFF
--- a/bld/hlpview/c/filelist.c
+++ b/bld/hlpview/c/filelist.c
@@ -187,7 +187,12 @@ static FileList *doFillFileList( const char *path, FileList *list )
         }
         *path_end = '\0';
         strcpy( buf, p );
-        len = strlen( buf ) - 1;
+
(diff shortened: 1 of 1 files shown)


jmalak:
The fix should provide a transition to processing the next segment if the segment is empty.
The proposed fix only fixes the crash, not the bug in the algorithm that caused the problem.

jmalak:
What about following solution
```
        if( *p != '\0' ) {
            strcpy( buf, p );
            len = strlen( buf ) - 1;
#ifdef __UNIX__
            if( buf[len] == '/' )
                buf[len] = '\0';
#else
            if( buf[len] == '/' || buf[len] == '\\' )
                buf[len] = '\0';
#endif
            list = scanDirectory( buf, list );
        }
```

jmalak:
fixed correctly in git repository

PR #1314, opened 2024-07-30
wdump: Add more type & flag infos for program & section headers

winspool:
Recognize more types for the ELF program and section headers,
Add missing infos for the flag field in the program headers and section headers

--
Regards ... Detlef


jmalak:
The GNU extension is not standard, so it should be clear that it is a GNU extension. For example, add "GNU - " before the existing text.

winspool:
> The GNU extension is not standard, so it should be clear that it is a GNU extension.
> For example, add "GNU - " before the existing text.

As suggested, i added a "GNU" prefix in the output

jmalak:
Thanks for your contribution.

PR #1315, opened 2024-08-02
aui_sample: Remove wrong Prototype

winspool:
This fixes a compile error on x86_64-linux

(More is needed for building on linux)

-- 
Regards ... Detlef


jmalak:
correct prototype already exists in app.h header file that this one should be removed.

jmalak:
Thanks for your contribution.

PR #1319, opened 2024-08-07
wdump: Use unions for program + section header

winspool:
Next step for 64bit ELF support 

--
Regards ... Detlef

### DIFF
--- a/bld/exedump/c/elfexe.c
+++ b/bld/exedump/c/elfexe.c
@@ -87,7 +87,7 @@ static  const_string_table elf64_exe_msg[] = {
     NULL
 };
 
-static  const_string_table elf_prog_msg[] = {
+static  const_string_table elf32_prog_msg[] = {
     "4type of segment                             = ",
     "4offset of segment from beginning of file    = ",
     "4segment virtual address                     = ",
@@ -99,7 +99,7 @@ static  const_string_table elf_prog_msg[] = {
     NULL
 };
 
-static  const_string_table elf_sec_msg[] = {
+static  const_string_table elf32_sec_msg[] = {
     "4name of the section                         = ",
     "4section type                                = ",
     "4flags                                       = ",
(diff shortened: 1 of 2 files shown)


jmalak:
Are you sure it is correct solution to mix 32-bit and 64-bit processing?
How many code can be share and how efficient it will be for other processed ELF info I don't mean only header.
By example for PE format it has sense because only part of header and a few items are different that shared code has sense but processing of ELF32 and ELF64 required much more items with different size.
Question is what is better if separated 32-bit and 64-bit processing or shared processing of both with many exceptions where it is different? I think you should have some idea if you are dealing with it now.

winspool:
I planed to use separate functions for dumping relocs and debug infos,
but i will try to do a 32bit/64bit split at the level of dmp_prog_sec

jmalak:
In this case I wouldn't complicate it by introduction of union when it is not necessary.
I am not much familiar with ELF format, if structure is same for 32-bit and 64-bit and difference is in fields size then could be possible to use 64-bit structure with 32-bit data that on begining copy 32-bit data into 64-bit structure and control it by flag if it is 32/64-bit. If unsigned_64 structure is used then you can use low 32-bit part simply and ignor high 32-bit for 32-bit ELF data in 64-bit structure. It is only quick idea I doin't know if it has sense.

winspool:
I already useed the following code, to manage the 32bit/64bit difference.
I will test, how using an Elf64 struct version works.

```
    int             e_phnum;
    int             e_shnum;
    unsigned        e_shstrndx;
    unsigned_32     sizeof_Phdr;
    unsigned_32     sizeof_Shdr;

    if( IS_ELF64( Elf_head) ) {
        e_phnum = Elf_head.elf64.e_phnum;
        e_shnum = Elf_head.elf64.e_shnum;
        e_shstrndx = Elf_head.elf64.e_shstrndx;
        sizeof_Phdr = sizeof( Elf64_Phdr);
        sizeof_Shdr = sizeof( Elf64_Shdr);
        offset = Elf_head.elf64.e_shoff
               + Elf_head.elf64.e_shstrndx * Elf_head.elf64.e_shentsize + start;
    } else {
        e_phnum = Elf_head.elf32.e_phnum;
        e_shnum = Elf_head.elf32.e_shnum;
        e_shstrndx = Elf_head.elf32.e_shstrndx;
        sizeof_Phdr = sizeof( Elf32_Phdr);
        sizeof_Shdr = sizeof( Elf32_Shdr);
        offset = Elf_head.elf32.e_shoff
               + e_shstrndx * Elf_head.elf32.e_shentsize + start;
    }
```

For other items, i get compile errors with `unsigned_64`.
I fixed that by using `uint64_t` in the Elf64 typedefs,
but last time, you introduced a new macro to make the code work with `unsigned_64`.

Where do you think, that `uint64_t` does not work, and using `unsigned_64`is required?
The OW source already needs C99.

jmalak:
Here is nice [comparison of 32 and 64-bit ELF](https://interrupt.memfault.com/blog/elf-format-differences#:~:text=ELF%2D64%20specifies%20byte%207,defined%20values%20for%20forward%20compatibility.)

jmalak:
It looks like following ELF structures have same structure that 64-bit variant, they could be used for 32-bit too:
File Header
Section Header
Relocation Entries

Following ELF structures have different structure and cannot be shared, it requires separate versions:
Symbol Table Entry
Program Headers

I don't know which structure you will use if it has sense.

winspool:
> Problem with unsigned_64 is that it is structure with union that you need address members instead of value,

```
xxx.u._64[0] is 64-bit value
xxx.u._32[0] is low 32-bit value
xxx.u._32[1] is high 32-bit value
xxx.u._16[0....3] is approriate 16-bit values
xxx.u._8[0....7] is appropriate bytes of value
```
Your hints above, how to correctly work with `unsigned_64` are really important.
They made it much easier for me to understand exactly, how `unsigned_64` works 

> exeelf.h uses unsigned_64 that you should use this type
> or elf types from exeelf.h which are based on unsigned_64.

> If you will share 64-bit ELF by 32-bit ELF data then it simplify access to 32-bit data 

I extended my test programm with a `unsigned_64` code path,
and learned a lot on my way to get it working.

PR #1322, opened 2024-08-14
clib/linux: Fix sysconf(_SC_NPROCESSORS_ONLN) result

winspool:
Needed to fix sysconf(_SC_NPROCESSORS_ONLN)

Small testcase:
```

#include <stdio.h>
#include <unistd.h>

#ifndef _SC_NPROCESSORS_ONLN
#error "_SC_NPROCESSORS_ONLN not found"
#endif

int main(void)
{
    long res = sysconf(_SC_NPROCESSORS_ONLN);
    printf("%ld\n", res);
    return 0;
}
```

The initialization of the 128 byte array increases the 
resulting program by 512 byte,
because OpenWatcom creates 128 mov instructions (even with owcc -Os`),
but that is a different issue.

--
Regards ... Detlef

### DIFF
--- a/bld/clib/linux/c/sysconf.c
+++ b/bld/clib/linux/c/sysconf.c
@@ -54,13 +54,20 @@
 static int __sysconf_nprocessors( void )
 {
     syscall_res res;
-    unsigned char mask[128];
+    unsigned char mask[128];    /* enough space for 1024 cores */
     int ret;
     int i;
+    int used;
 
     res = sys_call3( SYS_sched_getaffinity, (u_long)0, (u_long)(sizeof(mask)), (u_long)mask );
+    if ( __syscall_iserror( res ) ) {
+        _RWD_errno = __syscall_errno( res );
+        return( -1 );
+    }
+    used = __syscall_val( int, res );
+
     ret = 0;
-    for( i = 0; i < sizeof( mask ); i++ ) {
+    for( i = 0; i < used ; i++ ) {
         while( mask[i] ) {
             mask[i] &= mask[i] - 1;
             ret++;


jmalak:
I don't understand what you are trying to fix.
Many unrelated confusing info, but no core problem description.
Why the array must be initialized?
It looks like symptom of different problem, but without problem description I can not accept something strange what probably mask different problem.

winspool:
> I don't understand what you are trying to fix.

`sysconf(_SC_NPROCESSORS_ONLN)`

> Many unrelated confusing info, but no core problem description.

The syscall returns different broken results.
Examples: 175, 163, 177, 171, 180
(With my fix, 12 is returned every time)

> Why the array must be initialized?

The number of "1" bits in the whole array are counted.
The kernel syscall does not set the unused memory in the buffer to 0.

> It looks like symptom of different problem, 

You can see with this diff, how that function works:
[0000_debug_sysconf_nprocessors.txt](https://github.com/user-attachments/files/16616343/0000_debug_sysconf_nprocessors.txt)

Example output (I even tried with different optimization level): 
[_out_nproc.txt](https://github.com/user-attachments/files/16616374/_out_nproc.txt)

> but without problem description

I named the broken funktion (`sysconf(_SC_NPROCESSORS_ONLN)`) and added a simple testcase.

Unfortunately, i forgot the example output,
but using the small testcase before and after my fix is easy.

jmalak:
but uninitialized array has nothing to do with problem.
It is wrong implementation of __sysconf_nprocessors which not handle any error etc.
I suppose that array should be zeroed not fill by 1's.

jmalak:
If I read info about kernel call sched_getaffinity then it should return size of array which is setup by this call.
That it should be checking only such range.

jmalak:

PR #1348, opened 2024-11-18
Define NT_TIB[32/64] types in winnt.h

ivmai:
NT_TIB, NT_TIB32, NT_TIB64 types should be defined in h/nt/winnt.h, duplicating the definitions in h/nt/ddk/ntddk.h.

### DIFF
--- a/bld/w32api/include/ddk/ntddk.mh
+++ b/bld/w32api/include/ddk/ntddk.mh
@@ -1067,50 +1067,7 @@ typedef struct _KEY_VIRTUALIZATION_INFORMATION {
 } KEY_VIRTUALIZATION_INFORMATION;
 typedef KEY_VIRTUALIZATION_INFORMATION  *PKEY_VIRTUALIZATION_INFORMATION;
 
-/* NT thread information block */
-typedef struct _NT_TIB {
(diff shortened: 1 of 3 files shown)


ivmai:
Sample usage of NT_TIB (in client code): https://github.com/ivmai/bdwgc/blob/2ea2415a60ccc3393d3d235f73b4cab6f6fa8221/win32_threads.c#L859
I've tested this PR in the given project.

jmalak:
Hi,
Thank you for your patch.
Please, rework it little bit following way.
create file bld/w32api/include/nttib.sp which will contain only NT_TIB related structure definitions which exist in bld/w32api/include/ddk/ntddk.mh and is now missing in winnt.h.
remove this info from bld/w32api/include/ddk/ntddk.mh and replace it by single line
:include nttib.sp
add same line to bld/w32api/include/winnt.mh on correct position

It will fix appropriate header files as you proposed but ensure that single definition exists in nttib.sp and it is included to both header files.

ivmai:
Okay

ivmai:
> Please, rework it little bit following way.

Done

jmalak:
Many thanks for your contribution.

PR #1353, opened 2024-11-24
endian.h: Make sure, that _ARCH_INCLUDE always works

winspool:
Without this fix, projects failed to compile, when "endian.h" is included early or alone.

Example: zlib-ng
https://github.com/zlib-ng/zlib-ng/blob/develop/zendian.h

(zlib-ng has portability issues, before it can be build with OpenWatcom)

#####

I checked with `git grep`, that all other header, which use `_ARCH_INCLUDE, already include "incdir.sp".
Either nearby or for "sys/socket.mh" at the start of the file.

-- 
Regards ... Detlef

### DIFF
--- a/bld/hdr/watcom/endian.mh
+++ b/bld/hdr/watcom/endian.mh
@@ -38,6 +38,8 @@
 )
 
 :segment LINUX
+:include incdir.sp
+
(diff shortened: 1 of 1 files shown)


jmalak:

PR #1364, opened 2025-01-01
Return _exception::retval on math errors from math functions

dmo2118:
In the back of the Open Watcom C Library reference, appendix A.4 Domain Errors, there's a [list of error results for math functions](https://github.com/open-watcom/open-watcom-v2/blob/e14ae58ab7059b64d8ca074814ca7c5090b5e637/docs/doc/lr/src/clibfunc.gml#L109). Trouble is, it doesn't seem to match reality.

Here's a program:

```c
#include <stdio.h>
#include <math.h>

static void _print(const char *prefix, double x)
{
	fputs(prefix, stdout);
	fputs(" = ", stdout);
	if(x == HUGE_VAL)
		fputs("HUGE_VAL\n", stdout);
	else if(x == -HUGE_VAL)
		fputs("-HUGE_VAL\n", stdout);
	else
		printf("%g\n", x);
}

#define PRINT(x) _print(#x, (x))

#if __WATCOMC__

int _no_matherr(struct _exception *exc)
{
	//PRINT(exc->retval);
	return 123;
}

#endif

int main()
{
#if __WATCOMC__
	//_set_matherr(_no_matherr);
#endif

	PRINT(acos(2));
	PRINT(acosh(0.5));
	PRINT(asin(2));
	PRINT(atan2(0, 0));
	PRINT(atanh(-2));
	PRINT(log(0));
	PRINT(log10(0));
	PRINT(log2(0));
	PRINT(pow(-1, 0.5));
	PRINT(pow(0, 0));
	PRINT(pow(0, -1));
	PRINT(sqrt(-1));
	PRINT(y0(-1));
	PRINT(y1(-1));
	PRINT(yn(2, -1));
	return 0;
}
```

But the results are all zero: (on Linux, from `owcc -Wall temp.c -o temp && ./temp`)

```
acos(2) = 0
acosh(0.5) = 0
asin(2) = 0
atan2(0, 0) = 0
atanh(-2) = 0
log(0) = 0
log10(0) = 0
log2(0) = 0
pow(-1, 0.5) = 0
pow(0, 0) = 0
pow(0, -1) = 0
sqrt(-1) = 0
y0(-1) = 0
y1(-1) = 0
yn(2, -1) = 0
```

And if I uncomment the `_set_matherr`, I get (almost) all `123`:

```
acos(2) = 123
acosh(0.5) = 123
asin(2) = 123
atan2(0, 0) = 0
atanh(-2) = 123
log(0) = 123
log10(0) = 123
log2(0) = 123
pow(-1, 0.5) = 123
pow(0, 0) = 123
pow(0, -1) = 123
sqrt(-1) = 123
y0(-1) = 123
y1(-1) = 123
yn(2, -1) = 123
```

The patch in this PR gets almost to what the manual says things should be:

```
acos(2) = 0
acosh(0.5) = -HUGE_VAL
asin(2) = 0
atan2(0, 0) = 0
atanh(-2) = -HUGE_VAL
log(0) = -HUGE_VAL
log10(0) = -HUGE_VAL
log2(0) = -HUGE_VAL
pow(-1, 0.5) = 0
pow(0, 0) = 1
pow(0, -1) = HUGE_VAL
sqrt(-1) = 0
y0(-1) = -HUGE_VAL
y1(-1) = -HUGE_VAL
yn(2, -1) = -HUGE_VAL
```

...with the only difference being the `pow(0, -1)`.

A few things:
* `builder test` fails on my end in `open-watcom-v2/bld`, but it does pass in `open-watcom-v2/bld/mathlib`...for what that's worth, since the `mathlib` test shows no output.
* After the patch, with `_set_matherr(_no_matherr)`, the `123` returned from `_no_matherr` is completely ignored.
* You can't (not previously, and not now) set `struct _exception::retval` to override the default error result, even though that should work. (One problem at a time!)


jmalak:
Thanks for your contribution.

PR #1369, opened 2025-01-07
Update bundled libraries

mal359:
Sybase OW development did this until it stalled out; probably a good idea to take a look here, since those that are from the 2000's are full of nasty vulnerabilities that could affect OW2.


jmalak:
Sorry, such "mega" change is not accepted, nobody will be checking this.
Please submit changes per project (each project as zlib or libzib etc.) because some changes can be disputted or refused.
It requires review and our resources are limited, it take a time.

PR #1394, opened 2025-02-08
add OpenBSD support

jg1uaa:
There is three commits.

1. Different from other *BSDs, OpenBSD does not have interface for obtaining filname that is currently running such as NetBSD's /proc/curproc/exe . I wrote a code that uses argv[0], realpath() and PATH environment variable to emulate that interface.
2. Unicode related code at bld/wipfc defines char16_t, but the definition is different from OpenBSD's one. This causes error. It might be avoid with -DU_HAVE_CHAR16_T at configure, but this is disabled by [configure.ac](https://github.com/open-watcom/open-watcom-v2/blob/master/bld/wipfc/configure.ac#L5). So, workaround for [wipfc/icu/common/unicode/platform.h](https://github.com/open-watcom/open-watcom-v2/blob/master/bld/wipfc/icu/common/unicode/platform.h#L752) is required. It is worth replacing with modernized ICU4C code in the future.
3. Some components use DOSBox/DOSBox-X to run docs/gml/dos/wgml.exe but this is too slow on NetBSD/OpenBSD. As DOSBox's CPU speed  default as `cycles=auto` but even if adding `cycles=max` worked slow, I put enough large value that is used [old 86Duino's IDE](https://github.com/roboard/86Duino/blob/Coding-106/build/linux/work/DOSBox-0.74/dosbox-86duino.conf#L29).

### DIFF
--- a/bld/watcom/c/clibext.c
+++ b/bld/watcom/c/clibext.c
@@ -888,7 +888,9 @@ char *_cmdname(char *name)
 
     save_errno = errno;
 
-    if ( ( *_argv[0] == '.' || *_argv[0] == '/' ) && !_cmdname_sub( name, NULL, _argv[0] ) ) {
+    if ( ( _argv != NULL && _argv[0] != NULL )
+      && ( *_argv[0] == '.' || *_argv[0] == '/' )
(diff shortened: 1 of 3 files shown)


jmalak:
Thanks for your contribution.

PR #1397, opened 2025-02-08
wtouch/c/touch.c: add _argc, _argv (clibint.h) setup

jg1uaa:
I felt curious _argv is NULL in wtouch, it is quite easy reason... it is not set.
Other modules set _argv and _argc, wtouch do same thing.


jmalak:
Thanks for fix it.

PR #1398, opened 2025-02-09
msgencod, optencod, errdoc, vicomp: fix segv without argument

jg1uaa:
in many cases, adding argv/argc check or NULL by fopen() solve the problem.
maybe default value of FILE * variables are NULL, but I clarified it.


jmalak:
Thanks for your contributions.

Issue #1254, opened 2024-03-14
Support for constants with binary prefix 0b or 0B is missing


winspool:
I tried to compile the capstone engine with OW,
but that fails with many lines of:
```
Error! E1163: Invalid octal constant
```

The affected code locations are not octal constants, but lists of values in `typedef enum`
and the values are binary constants.
Examples
```
  ARM_T = 0b1000
  PPC_BO_TEST_CR = 0b10000
```

###

binary constants / binary literals with the binary prefix "0b" or "0B" are in the standards:

For C23, the requirement is listed in section 6.4.4.1 Integer Constants
For C++14, the requirement is listed in section 2.13.2 Integer literals [lex.icon]

PR #1410, opened 2025-02-18
cc: Add support for binary constants [C23] (#1254)

winspool:
Changes in several areas for the implementation are included:
```
cc/c/coptions.c:  Accept -zastd=c23
cc/c/cscan.c:     Read binary constant with a "0b" or a "0B" prefix
cc/gml/cerrs.gml: New Error Message: E1208: Invalid binary constant
watcom/?/i64.?:   New Helper function: U64Cnv2
```

New tests are included:
```
positive/*/binary01: Binary constants, tested in C23 mode, without extensions
positive/*/binary02: Binary constants, tested in C99 mode, with extensions
diagnose/*/diag0103: Binary constants, tested in default mode, without extensions
diagnose/*/diag0104: Bad binary constants, tested in default mode, with extensions
```

The C23 feature to make [long] numbers more readable with ' characters
in the number is not implemented.

--
Regards ... Detlef


jmalak:
Thanks for your contribution.
It looks OK.
It is necessary include this feature conditionally for standards lower then C23 as extension.
Add also regression tests for this new feature to bld/ctest/diagnostics and bld/ctest/positive.

winspool:
I rebases the implementation.

For the tests, this is the current patch status:    
[0002-bld-ctest-Add-tests-for-binary-constants.txt](https://github.com/user-attachments/files/19035400/0002-bld-ctest-Add-tests-for-binary-constants.txt)

Are these tests fit, what you suggested?

jmalak:
anyway why you don't follow evident rules for test files numbering and introduce something chaotic. 
In OW all has own purpose and is designed to handle it properly if not then it is time to change.
This is not candidate for change.

jmalak:
next note: why you test it with **`-za99`** if it has no relation to **`C99`** standard, it is extension for every C standard version below **`C23`**.
you can simply test it without **`-za99`** or use **`-zastd=c23`** if need test it with this option.
For switch extensions on/off is options -za and -ze.
**`-za99`** is logical combination of C standard switching and -za option.
**`-za99 -ze`** is C99 standard with extension

winspool:
> next note: why you test it with **`-za99`** if it has no relation to **`C99`** standard.
> This feature is extension for every C standard version below **`C23`**.
> you can simply test it without **`-za99`** or use **`-zastd=c23`** if need test it with this option.

I used c99 as the lowest std,
because the test uses 64bit int values,
which were added for c99:
One positive test is uses c23 (without extension)
The other positive test is uses c99 (for int64_t and  uint64_t)
with extensions.

> **`-za99`** is logical combination of C standard switching and **`-za`** option.

I can't see that in coptions.c
I found only, that `-za99` sets cstd to STD_C99

The reworked patch will include the implementation and the tests
and i will recheck the options.

jmalak:
Sorry, you are right the any -zann or -zastd=cnn doesn't change status of language extensions.
lowest C standard is C89 and it is default not need to setup only -ze and -za is OK to switch extension on/off.
that test should cover combination -za or -zastd=c23 + -za or no options I think.

jmalak:
Note for tests.
take into account that compiler is optimizing that expression where you add constant 0 is useless because it is optimize out by compiler until you disable optimization.
You should always use in such expression non zero constant to compiler process constant in generated code and doesn't optimize it out.
best is to check generated code if compiler realy do what is expected because optimization can optimize out constants or constant expression.
Please rework tests a little bit to work reliably and realy not optimize out constants then test has no sense.

PR #1413, opened 2025-02-27
bld/wstuba: add BSD support

jg1uaa:
binb64/wstub.exe not exist, fixed.


jmalak:
Thanks for your contribution.

PR #1442, opened 2025-03-31
cc: Fix detection of -zastd=c17

winspool:


pchapin:
This looks like a correct fix to me. I don't know why the CI build is unhappy. @jmalak, do you have any thoughts?

pchapin:
Thanks for your contribution!

PR #1447, opened 2025-04-05
wpp*: Fix handling of the #warning preprocessor directive

winspool:
The previous code enabled the preprocessor directive "#warning",
when extensions are disabled (independent of the selected c++ standard).

This PR fixes this.

### DIFF
--- a/bld/plusplus/c/cmac2.c
+++ b/bld/plusplus/c/cmac2.c
@@ -700,9 +700,9 @@ static void CWarning( void )
 {
     bool save;
 
-    if( !CompFlags.extensions_enabled ) {
+    if( CompFlags.extensions_enabled || ( CHECK_STD( >= , CXX23 ) ) ) {
         get_arg_message();
-        /* Force #error output to be reported, even with preprocessor */
+        /* Force #warning output to be reported, even with preprocessor */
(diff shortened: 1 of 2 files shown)


winspool:
Extending the code to handle "-zastd=c++**" in the same way as currently used
is possible:
[0000_wpp_recognize_more_standards.txt](https://github.com/user-attachments/files/19612256/0000_wpp_recognize_more_standards.txt)

With a table, the code might look cleaner, but how to handle that
with the commandline handling macros?

pchapin:
While we're here, I suggest we add support for a /zastd=c++03 option, which is a synonym for c++98. I'm not sure if other compilers do that, but it should be straightforward, so why not?

winspool:
> While we're here, I suggest we add support for a /zastd=c++03 option, which is a synonym for c++98.
> I'm not sure if other compilers do that, 

clang and gcc do it
> but it should be straightforward, so why not?

This patch contains only a small code change to fix the handling of `#warning`.
A patch for all c++ standards is available (needs a small cleanup):
https://github.com/user-attachments/files/19639265/0001-plusplus-Detect-all-c-standards-use-them-for-__cplus.txt

A comment from @jmalak, what he want's, would be helpful.

PR #1450, opened 2025-04-13
plusplus: Detect all c++ standards, fix "#warning" availability

winspool:
Now "#warning" is available in c++23
and also as extension in older C++ standards.


bld/plusplus/c/cmac2.c
jmalak, line 1: Should use new solution, not old

PR #1452, opened 2025-04-24
owcc: Overwrite the default program name only for unix targets

winspool:
When cross-compiling on linux for any Windows (or DOS, or OS/2) target,
OpenWatcom now creates a program with an ".exe" extension by default.

Any autotools project now detects the ".exe" extension 
when cross-compiling with OpenWatcom
and uses the ".exe" extension for creating the project executables.

#####

This is a usability enhancement, which makes OpenWatcom more usable
for cross-compiling any autotools project.
In addition, now `owcc -bnt helloworld.c` and `wcl386 -bcl=nt helloworld.c`
produce both a `helloworld.exe` on linux.

### DIFF
--- a/bld/wcl/c/owcc.c
+++ b/bld/wcl/c/owcc.c
@@ -1514,12 +1514,28 @@ static  int  CompLink( void )
                 rc = tool_exec( TYPE_DIS, ofname, dis_args );
             }
             if( Exe_Name == NULL ) {
+                /* overwrite the default program name only on unix targets */
+                if( Flags.link_for_sys ) {
+                    /* check, if the target system is a unix system */
+                    if( strncmp( SystemName, "linux", 5 ) == 0 ) {
+                        Flags.keep_exename = 1;
+                    } else {
+                        Flags.keep_exename = 0;
+                    }
+                } else {
+                    /* called without "-bsystem" argument. use the host to decide */
 #ifdef __UNIX__
-                Exe_Name = MemStrDup( OUTPUTFILE );
-                Flags.keep_exename = 1;
+                    Flags.keep_exename = 1;
 #else
-                Exe_Name = MemStrDup( RemoveExt( Word ) );
+                    Flags.keep_exename = 0;
 #endif
+                }
+                if( Flags.keep_exename ) {
+                    /* use the well known default program name on unix targets */
+                    Exe_Name = MemStrDup( OUTPUTFILE );
+                } else {
+                    Exe_Name = MemStrDup( RemoveExt( Word ) );
+                }
             }
             file = GetName( NULL, NULL );   /* get next filename */
         }


jmalak:
Please give a reasonable reason why we should break backward compatibility for all projects using the owcc tool.

winspool:
autotools creates a "configure" script,
which detects the default executable extension and the default program name.
(`$CC  $CFLAGS conftest.c`) Detection is only possible without any "-o" option. 

These infos are then uses to build the project.
When using `owcc` on linux to cross-compile for Windows (or DOS),
"a.out" is found in the detection phase
and the programs of the project are created without an "exe" extension.

Automatic testing the build results or preparing an installer do not work with `owcc`.
All other cross-compile toolchains create a "ready to use" program with an exe extension.

Cross compiling with `wcl386 -bcl=nt` also produces a program with an exe extension.

----

Whith my change, using `owcc` for `$CC  $CFLAGS conftest.c` creates `conftest.exe`
and the resulting project programs have the expected exe extension.


Related configure script detection steps:
* checking whether the C compiler works
* checking for C compiler default output file name
* checking for suffix of executables

jmalak:
I don't see any reason why we need to broke OW backward compatibility and you didn't give me any.
using option -o is proper solution, if autotools is not capable to use it then it is autotools issue not OW.
OW will never be compatible with GNU tools, so similar requests that would break OW's backward compatibility will not be accepted.

winspool:
> I don't see any reason why we need to broke OW backward compatibility and you didn't give me any.

A less invasive option can be to change the default output name only,
when owcc is running on a unix systems
and keep current default names, when owcc runs on Windows (or DOS, or OS/2)

> using option -o is proper solution, if autotools is not capable to use it then it is autotools issue not OW.

autotools runs the compiler frontend without the "-o" option to detect the default executable name and the default extension.
(After the detection phase, the "-o" option is used.)

Since current OpenWatcom creates an "a.out",
when compiling on a unix host for a Windows (DOS, OS/2,...) target,
autotools finds the "a.out" without ".exe" extension
and this results in all programs in the project to be  build without the ".exe" extension.

> what about owcc object file extension which is also different for *NIX platforms.

autotools handles ".o" as well as ".obj" extensions for object files.
(Probably more. I'm currently not at home)

jmalak:
GNU tools are switch by target (you need to install and configure for each target), but OW doesn't use such logic. OW is native cross-compile tools that you have one tool for all targets and therefore default are per host system not for target.
As usualy you want GNU behaviour for OW tools but they use absolutely different logic. You need to change autotools not OW.

jmalak:
Your issue can be resolved simply by use OWCC environment variable to include default option which you want to run owcc utility. These options are included in front of command line options when owcc start.

PR #1476, opened 2025-07-03
clib: fix NULL pointer dereferences

stsp:
If /etc/hosts or /etc/services is missing, the corresponding functions gethostent() and getservent() crash.
This patch avoids the problem by ading the missed NULL checks.

Note: the patch is trivial, but it was untested on the main OW source tree. The testing was only done on a local rip-off.


jmalak:
Thanks for your contribution.

PR #1479, opened 2025-07-10
Dns

stsp:
With these small changes I've
got DNS resolver to work.

Not tested on the OW build -
I use my own rip-off, although
the changes look rather simple.


jmalak:
Thanks for your contribution.

PR #1480, opened 2025-07-15
resolver: fix leak of query_socket

stsp:
Not tested in OW code base.

### DIFF
--- a/bld/clib/posix/c/_resolve.c
+++ b/bld/clib/posix/c/_resolve.c
@@ -357,6 +357,11 @@ int _dns_query( const char *name, int query_type, in_addr_t dnsaddr, struct host
 
 dns_cleanup:
 
+#if defined( __LINUX__ )
+    close( query_socket );
+#else
+#error please provide closesocket( query_socket ) for your OS
+#endif
     if( answers != NULL && dns != NULL ) {
         for( i = 0; i < n_answers; i++ ) {
             if( answers[i].name != NULL )


jmalak:
Thanks you for your contribution.

It should be fixed to use only
**`close( query_socket );`**
because it is implemented for Linux only (now).
Other OS-es has different socket functions that it will require implement it little bit different way.
Socket API for RDOS is very specific that it will need much more changes.
OS/2 by example use **`soclose`** Windows **`closesocket`** etc.
You can look for [TCP trap code](https://github.com/open-watcom/open-watcom-v2/blob/master/bld/trap/tcp/c/tcplink.c) to see how much different are socket API for various OS-es.

For now use only Linux code.

stsp:
If not for the fact that I
myself am testing this not
under linux. :)
So done now.

rdos314:
RDOS use it's own trap files and not the ones in OW. This is partly due to differences in the socket interface, but also differences in how debugging itself works. In RDOS, only the debugged thread is stopped, not the whole application, and you can also trace into kernel. This made it necessary to build custom debug servers.

stsp:
What is RDOS, where are its
sources on github?

rdos314:
It's not yet on github, rather is only available with SVN on my site: https://www.rdos.net/rdos. I have changed the license to MIT now and will soon move it to github to be able to build it automatically, but also to easier being able to integrate with various software here.

stsp:
```
svn: E175003: The server at 'https://svn.rdos.se/trunk' does not support the HTTP/DAV protocol
```

Tried http instead of https:
```
$ svn co  http://svn.rdos.se/trunk
svn: E170013: Unable to connect to a repository at URL 'http://svn.rdos.se/trunk'
svn: E000002: Could not find the requested SVN filesystem
```

Hmm.

rdos314:
This link works for me: http://svn.rdos.se/rdos/trunk

My site use Ubuntu, but I was unable to add two different SVNs and https.

stsp:
Yes, that works, thanks.
So you have a bad URL here:
https://www.rdos.net/rdos/
Which says `http://svn.rdos.se/trunk`
and that doesn't work.
So it seems it can emulate DOS
and run dos progs?

rdos314:
Fixed the link and updated the license info.

While rdos initially was aimed to emulate DOS, and run dos progs (in 1988), this support is likely broken now as this has not been tested or used for a long time. RDOS is a completely new design that fixed many of the problems with DOS, like the lack of multitasking. It also supports SMP now and can boot using 32- and 64-bit EFI into 32-bit protected mode or through legacy boot.

stsp:
Maybe you can re-create DOS
support with this library:
https://github.com/dosemu2/fdpp
Although it uses libelf and needs
some unix tools for build, so for
RDOS it probably isn't going to be
easy (but eg Haiku builds it w/o any
problems).

PR #1481, opened 2025-07-15
resolver: h_addr_list must contain ptrs to struct sockaddr

stsp:
Not a string representations.
Therefore instead of using inet_ntoa(), we actually need to memcpy the raw sockaddr structs.

Not tested in OW code.


stsp:
Example:
https://gist.github.com/listnukira/4045628#file-get_ip_v2-c-L30

jmalak:
Thanks for your contribution.

I think that **`_add_string_to_list`** should be renamed to something like **`_add_item_to_list`**.
Now it is confusing.

stsp:
Renamed.

PR #1484, opened 2025-07-27
resolve: reduce DNS buffer size

stsp:
This patch is needed for a real-mode
port of that code.
Not tested in OW tree, as it doesn't
build this for real-mode.


jmalak:
Thank you or your contribution.

PR #1492, opened 2025-08-11
For 16-bit windows, force __doserror_ as near intra-segment symbol to avoid FAR CALL instruction

joncampbell123:
Real-mode Windows, for MOVEABLE segments, redirects FAR calls through an export table entry that starts with a "SAR" instruction followed by a FAR JMP. The SAR (shift arithmetic right) is apparently how the Windows LRU algorithm determines which segments are most often used so that it can discard segments not in use, however it also corrupts the CF (carry) flag which _doserror needs to do it's job properly. As a near intra-segment symbol, the real-mode Windows FAR JMP and tracking is eliminated and the carry flag is preserved.

This patch only affects the C runtime for 16-bit Windows, all MS-DOS runtimes are unaffected by this patch by design.

\_\_doserror\_ needs to act on the CF carry flag provided by the caller. This change allows open() to work with O_CREAT|O_TRUNC flags even in real-mode and create the file if it does not exist.


jmalak:
I think is is not correct solution for any BIG CODE model because any near function cannot be used globaly only copy of it per each module, because each module has its own far segment. Even if OW linker shrink segments to automatic segment group then it can work for small application, but in case group is over 64 kbytes then issue is evident because you have multiple group (far segments). The righ solution is to change _doserror_ function to not relate to Carry flag but pass Carry flag as parameter and use distance for appropriate memory model as now. I think it is mistake in implementation of _doserror_ for 16-bit Windows and must be fixed

jmalak:
I think is is not correct solution for any BIG CODE model because any near function cannot be used globaly only copy of it per each module, because each module has its own far segment. Even if OW linker shring segments to segment group then it can work, but if such group is over 64 kbytes then issue is evident because you have multiple group (far segments). The righ solution is to change _doserror_ function to not relate to Carry flag but pass Carry flag as parameter. I thinks it is mistake in implementation of _doserror_ for 16-bit Windows.

joncampbell123:
Should it PUSHF and then read from stack then? Or PUSHF / POP BX and read from BX? Or...?

joncampbell123:
Even in the BIG CODE model, the C runtime library is small enough to have it's own code segment all to itself, and \_\_doserror\_ is an internal function, so why not just call it near? You're not supposed to call \_\_doserror\_ outside the C runtime, correct?

jmalak:
You are right, if it is limited to use only for CRTL code, then it can be defined and called as a near function, since the CRTL code is in one near segment.
There is no need to define it separately for different memory models.
That solution should be simple only change declaration of _doserror to be near and use near call averywhere (it should be automaticaly by definition as near).

For assembler there is macro defn instead of defp and xrefn instead of xrefp.

joncampbell123:
I see. defn will guarantee it is a near procedure and the "ret" will be a near ret"?

I had to put __near around the header definition of doserror or else everyone else would call it as a FAR procedure even if it was compiled as near.

jmalak:
You need to change xrefp macro to xrefn that assembler know _doserror as near function.
ftime086.asm		bld/clib/_dos/a
ftime386.asm		bld/clib/_dos/a
io086.asm		bld/clib/_dos/a
io386.asm		bld/clib/_dos/a
iow32.asm		bld/clib/_dos/a
mem086.asm		bld/clib/_dos/a
mem386.asm		bld/clib/_dos/a

joncampbell123:
There is code that includes doserror.h and calls it like a C function. If the changes to doserror.h are not applied, some C code that calls it such as mkdir(), rmdir(), rename(), unlink(), etc will call it FAR incorrectly.

joncampbell123:
xrefn? xrefn? There's no such macro. There is xrefp, though the macro doesn't seem to differentiate between BIG CODE and otherwise.

EDIT: There is an xrefn now, in the pull request.

jmalak:
Please remove mdef.inc from PR because xrefn macro was already added.
Also rebase PR to the latest source

PR #1494, opened 2025-08-15
Win386 extender: Zero ES, FS, and GS segment registers before freeing DPMI alias LDT descriptor to avoid mysterious Windows 3.0 DPMI bug that causes Windows 3.0 to crash dump back to DOS within initialization stage

joncampbell123:
Way back when most DOSLIB code was written, Win386 code worked perfectly fine on Windows 3.0.

Somewhere along the way, Win386 executables started causing Windows 3.0 to crash dump back to DOS.

Through careful testing with the DOSBox-X debugger, somehow, it was found the DPMI call to free an alias LDT descriptor was causing it. Not by the function itself, but within the Windows DPMI server when INT 31h is called. As it turns out, zeroing the ES, FS, and GS registers prior to the DPMI call prevents the crash.

Compiled Win386 executables once again run in Windows 3.0 without crashing.

I know the target of Open Watcom is Windows 3.1 or higher, but I read that Watcom C made Win386 work at a time when most customers were still running Windows 3.0, and it did work in the past just fine back when I was writing most DOSLIB code around 2014 or so.


jmalak:
Thanks for your contribution. 
Any problem fixing is welcome even if it is not for latest Windows version.

Generaly **`_asm`** statement should not be used with OW code and **`#pragma aux`** should be used.
**`_asm`** construct should be replaced by **`#pragma aux`** which put complete info about changed registers to code generator which can take it into account.

```
extern void Clear_ES_FS_GS( void );
#pragma aux Clear_ES_FS_GS = \
    "xor ax,ax" \
    "mov es,ax" \
    "mov fs,ax" \
    "mov gs,ax" \
__modify __exact [__ax __es __fs __gs]
```

better should be to change **`DPMIFreeLDTDescriptor`** to do this regular way.
Now it is fixed on single place only but **`DPMIFreeLDTDescriptor`** is used in different locations.
I think we should create modified **`DPMIFreeLDTDescriptor`** for 16-bit Windows only and use it in WIN386 extender code.

joncampbell123:
Well, yes. The reason I didn't put that in the DPMIFreeLDTDescriptor was out of concern that some 286-level code might use it. The 286 does not have FS and GS, those didn't appear until the 386.

Perhaps you should name them DPMIFreeLDTDescriptor286 and DPMIFreeLDTDescriptor386 for 286 and 386 specific versions now that segment registers are involved.

joncampbell123:
If there is a define macro or ifdef that the code can use to know if the 16-bit compiler is targeting the 386 or not, it could auto define which one is the DPMIFreeLDTDescriptor function

jmalak:
OW 32-bit Windows extender can work only on 80386 and above by design, that it has no much sense to take into account 80286 for code.
If we talk about DPMI inline code then it should work on 80286.
I am preparing a few auxiliar function (80286) to support 16-bit DPMI environment.
Anyway I don't want touch DPMI inline code for DOS which is in dpmi.h and do copy of used DPMI functions in 32-bit extender to extender codebase where it can be fix or we can do override for a few functions.

joncampbell123:
OK. Let me know when I should close this pull request upon completion.

joncampbell123:
I made a change to zero registers like you suggested.

PR #1496, opened 2025-08-17
Watcom linker support for conditional blocks (if/elseif/else/endif), allow OPTION NODEFAULTLIBS to disable Windows 1.0 system block startup and alias linker directives

joncampbell123:
Watcom linker: Add support for if/elseif/else/endif to linker statement parser and make some specific key handlers take the if statement block into consideration. Add if statement to wlink.lnk so that the windows1 system blocks do NOT include default lib path or alias statements if OPTION NODEFAULTLIBS is given before invoking the system block. If someone wants to write bare metal ASM for Windows 1.0 they should be able to do it with OPTION NOSTDLIBS without issue.


joncampbell123:
Most of the code beyond the parser changes are kind of a hack to make it work and only a very specific statements obey the if conditions block, enough so that the system block for windows1 can contain an if statement based on OPTION NODEFAULTLIBS. Feel free to clean up or suggest any cleanups I can do.

joncampbell123:
Another possible future change is to allow parenthesis after if/elseif to allow compound conditions.

    if (defaultlibs and whatever)
       ...
    endif

    if ((defaultlibs and whatever and whateverelse) or somethingelse)
       ...
    endif

maybe even

    if (system is windows1 or system is windows1_dll)
       ...
    endif

Sound like a good thing for me to add in the future?

PR #1497, opened 2025-08-17
Watcom linker: If the target symbol of the ALIAS directive does not exist, and nobody uses the ALIAS symbol, do not report as missing symbol

joncampbell123:


jmalak:
I need to review it in more details, I think that it could be resolved by weak symbol attribute.
I don't know how alias symbols are resolved now.

joncampbell123:
OK. Test case DOSLIB tinyw16/hellowin. Prior to this change the ALIAS directives prevent it from linking due to missing symbols that nobody uses.

https://github.com/joncampbell123/doslib/tree/master/tinyw16/hellowin

PR #1501, opened 2025-08-22
resolver: initialize h_length

stsp:
It was left uninitialized.
The simplest fix, but the difficult to debug one.


jmalak:

PR #1508, opened 2025-09-07
cc: Support 64bit variables for switch [C99]

winspool:
Unfortunately, 64bit case values are still not supported.

They require also changes in the code generator and the code generator API.

--
Regards ... Detlef

### DIFF
--- a/bld/cc/c/cstmt.c
+++ b/bld/cc/c/cstmt.c
@@ -1028,6 +1028,12 @@ static void SwitchStmt( void )
     case TYP_LONG:
 //        switch_type = TYP_LONG;
         break;
+    case TYP_ULONG64:
+        sw->case_format = "%llu";
+        break;
+    case TYP_LONG64:
+        sw->case_format = "%lld";
+        break;
     default:
         CErr1( ERR_INVALID_TYPE_FOR_SWITCH );
     }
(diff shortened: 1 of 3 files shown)


jmalak:
Such change is absolutely useless.
At minimum it should implement code generator functions
**`void CGSelCase64( sel_handle s, label_handle lbl, signed_64 val )`**
**`void CGSelRange64( sel_handle s, signed_64 lo, signed_64 hi, label_handle lbl )`**
see [code generator documentation](https://open-watcom.github.io/open-watcom-v2-wikidocs/cgdoc.html) for 32-bit versions.
also **`void CGSelectRestricted( sel_handle s, cg_name expr, cg_switch_type allowed )`** will need some changes because not all methods of generating code is possible with 64-bit argument on 32 or 16-bit CPU.

Anyway it should not be PR but draft (WIP) untill it will be in usable status.

jmalak:
I am just working on code generator to support 64-bit values instead 32-bit values only.
I extended CG internal structures variable for values to be 64-bit and changed appropriate processing to be fully 64-bit aware.
As soon as I finishing with CG changes then Compiler front-ends can be changed to support 64-bit values (C, C++ and Fortran compiler), but it is much simpler then CG changes.

winspool:
Thanks.
I'm on holiday now and will look at it, when i'm back.

jmalak:
now 64-bit items support is in OW, that it can be tested.

PR #1571, opened 2026-01-07
fpe87: fix memory corruption in FPE handler

stsp:
When int2 handler was switching to FPEStk, the wrong stack segment was used: SI was the old segment, and we need new one, which is in BX. As the result, the stack was pointing to some unrelated "innocent" memory area, corrupting it.


jmalak:
Thanks for your contribution.
It is very very old bug in 16-bit C run-time library.

stsp:
Its probably even dated
to watcom v1. I remember
I usually had to reboot the
machine if some OW-based
game crashed with FPE, as the
DOS was no longer reliable.
Which was like 30 years ago.

andrewbird:
Thanks for this. When can I expect it to land in the github action you publish for Watcom?

jmalak:
now we have issue with dosbox-x on macOS that OW build is failing.
As soon as this issue will be fixed then OW build will be available

jmalak:
@andrewbird 
I finally managed to find a way to install an older version of dosbox on macOS ARM64 and the OW build is successfully created again. Tomorrow morning should be available new OW build with your fix.

andrewbird:
Lovely, I'll look forward to it.

andrewbird:
@jmalak Thanks, the action is working a treat now!

PR #1583, opened 2026-02-14
Fix is.../to... usage

cmeerw:
see https://man.netbsd.org/NetBSD-10.1/ctype.3

The argument of these functions is of type int, but only a very restricted subset of values are actually valid.  The argument must either be the value of the macro EOF (which has a negative value), or must be a non-negative value within the range representable as unsigned char.  Passing invalid values leads to undefined behavior.


jmalak:
Thanks for your contribution.

PR #1584, opened 2026-02-21
strnicmp and _fstrnicmp return incorrect value when n == 0

davidebreso:
When the comparison length parameter `n` is set to `0`, the functions `strnicmp` and `_fstrnicmp` return an incorrect, arbitrary non-zero value instead of indicating equality.

When `n == 0` the local variables `c1` and `c2` are left unitialized, and the arbitrary value `c2 - c1` is returned.
To correct the bug, return `c2 - c1` only when a mismatch occurs, `0` otherwise.


jmalak:

PR #1592, opened 2026-03-29
cc, plusplus: add -pm option to dump predefined macros

playday3008:
## Summary

Add a new `-pm` compiler option to both the C and C++ compilers that dumps all predefined macros in `#define` format to stdout (or to a file via `-fo`), similar to `gcc -dM -E` / `clang -dM -E`.

This is useful for inspecting which macros the compiler defines for a given target, CPU, memory model, and set of flags — something that was previously not possible without manually consulting the source.

`-pm` is not part of the `-p` preprocessor option chain (uses `:nochain.`/`:usagenochain.`), is processed after the `-p` group with highest priority, and does not require a source file.

### Usage

```sh
wcc386 -pm                 # 32-bit x86 C compiler
wcc -pm                    # 16-bit x86 C compiler
wcc386 -pm -bt=linux       # with target system
wpp386 -pm                 # 32-bit x86 C++ compiler
```

### Example output

```c
#define _M_IX86 300
#define _M_I386
#define __386__
#define __X86__
#define _X86_
#define __WATCOMC__ 1300
#define _PUSHPOP_SUPPORTED
#define __STDC_NO_VLA__
#define __WATCOM_INT64__
#define _INTEGRAL_MAX_BITS 64
#define _STDCALL_SUPPORTED
...
```

### Implementation

- Walks the compiler's internal macro hash table after command-line processing and prints each non-hidden, non-special macro entry
- Reuses the existing preprocessor output file mechanism (`CppFile`), so `-fo` works as expected
- Hidden macros (`__func__`) and special dynamic macros (`__LINE__`, `__FILE__`, `__DATE__`, `__TIME__`) are excluded, matching `gcc -dM` behavior
- C++ compiler uses `RingIterBeg`/`RingIterEnd` for its ring-based hash table

### Files changed

**C compiler (`bld/cc/`)**:
- `gml/options.gml` — option definition
- `h/ctypes.h` — `cpp_dump_macros` flag in `comp_flags`
- `c/coptions.c` — option handling
- `c/cmac1.c` — `DumpMacroDefn()` + `DumpAllMacros()`
- `h/cvars.h` — declaration
- `c/ccmain.c` — integration in `DoCCompile()`

**C++ compiler (`bld/plusplus/`)**:
- `gml/options.gml` — option definition
- `h/compflag.h` — `cpp_dump_macros` flag
- `c/cmdlnany.c` — option handling
- `c/cmacadd.c` — `dumpMacroDefn()` + `DumpAllMacros()`
- `h/preproc.h` — declaration
- `c/plusplus.c` — integration in `doCCompile()`

---

> This implementation was developed with the help of an LLM (Claude Code). All changes have been manually reviewed, verified, and tested by the author.

### DIFF
--- a/bld/cc/c/ccmain.c
+++ b/bld/cc/c/ccmain.c
@@ -1020,23 +1020,22 @@ static void DoCCompile( char **cmdline )
     }
     ParseInit();
     if( ParseCmdLine( cmdline ) ) {
-        if( WholeFName == NULL ) {
+        if( CompFlags.cpp_dump_macros ) {
+            if( CppFile == NULL )
+                OpenCppFile();
+            DumpAllMacros();
+            MacroFini();
+        } else if( WholeFName == NULL ) {
             CErr1( ERR_FILENAME_REQUIRED );
             return;
-        }
-        MakePgmName();
-        DelErrFile();               /* delete old error file */
-        CPragmaInit();              /* memory model is known now */
+        } else {
+            MakePgmName();
+            DelErrFile();               /* delete old error file */
+            CPragmaInit();              /* memory model is known now */
 #if _CPU == 370
-        ParseAuxFile();
+            ParseAuxFile();
 #endif
-        if( CompFlags.cpp_mode ) {
-            if( CompFlags.cpp_dump_macros ) {
-                if( CppFile == NULL )
-                    OpenCppFile();
-                DumpAllMacros();
-                MacroFini();
-            } else {
+            if( CompFlags.cpp_mode ) {
                 PrintWhiteSpace = true;
                 if( ForcePreInclude != NULL ) {
                     if( CppFile == NULL )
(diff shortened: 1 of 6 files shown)


jmalak:
Thanks for your contribution.

The -pm option need to be little changed.
The pre-processor control options start with p character and can be chained in p options group by leading p, but proposed option -pm is not part of pre-processing control group.
You can get idea from -pil option which use similar logic (not included in P chain) as proposed -pm, but it is here for backward compatibility. In processing you need take into account order of options on command line and code in which order is option processed if there are dependency on compiler internal flags. Probably -pm should be processed as last due to highest priority and overwrite some internal flags setup by -p pre-processor control group options.

jmalak:
@playday3008 
I just submit fix for null device as input device that now null device can be used for both C and C++ compiler.

PR #1594, opened 2026-03-31
wasm: treat Ctrl-Z as DOS text EOF

ddanila:
## Summary

WASM's source line reader does not recognize `0x1A` (Ctrl-Z, the DOS text end-of-file marker) as end-of-file. When an assembly source ends with that marker, the byte is forwarded to the scanner as a standalone token and can trigger an internal assertion instead of cleanly ending input.

This came up while assembling the preserved MS-DOS 4.0 source tree, specifically around the `CMD/DEBUG/DEBUG.ASM` build path.

## Changes

- treat `0x1A` as EOF in `get_asmline()` when it appears outside quoted text
- keep quoted literals unchanged; the fix is scoped to the unquoted DOS text-EOF case that triggered the crash
- add a `ctrlzeof` diagnostic regression case
- include the new regression in the `diag` WASM test set

## Impact

Assembly sources with a DOS Ctrl-Z terminator now stop at logical end-of-file instead of exposing a bare SUB byte to later parsing stages. This change is intentionally narrow: it applies to the source reader's unquoted text path and does not reinterpret `0x1A` inside quoted literals.

## Validation

- verified the Ctrl-Z reproducer directly
- ran the standalone WASM suite after a clean rebuild: `test1`, `test2`, `test3`, `diag`, and `ideal`


jmalak:
The Ctrl-Z is equal EOF without any relation to context as quoted string.
If Ctrl-Z is in middle of quoted string then it is user mistake and EOF is in middle of quoted string that rest of file must be abandoned. I thing it is from CP/M era.

ddanila:
> The Ctrl-Z is equal EOF without any relation to context as quoted string. If Ctrl-Z is in middle of quoted string then it is user mistake and EOF is in middle of quoted string that rest of file must be abandoned. I thing it is from CP/M era.

Absolutely. But such a file never would be compilable on any [such] systems I assume, I've just mentioned that to describe that this is intentionally left out from the scope of the given PR. This PR is to make "correct" files with explicit trailing EOF symbol to be compilable.

jmalak:
Normally Ctrl-Z is last character in file with DOS, but files from CP/M era can have Ctrl-Z earlier then on end. It is related to history of recording. Correct solution is replace Ctrl-Z by null terminator in reading buffer and stop next reading. Anyway source file should not contains control characters except a few one. Ctrl-Z is not regular character in text file.

jmalak:
I think there is simpler solution.
Because we use read only file access by getch (with buffering) then when Crtl-Z is reeded we can seek to file physical end and put null character in line buffer. Next read will return EOF and file will be closed. No need any flag or some initialization or algorithm change etc. Take into account that it must work on every used text file (included).

ddanila:
> I think there is simpler solution.

I was weighing exactly between these two solutions, and picked the previous one as "more proper". But looks like it does not ) Pushed another change.

jmalak:
There is no random access only sequential that no worry about issue even if looks strange, but it is simple and reliable, encapsulated into get_asmline (single access to the input file).

PR #1595, opened 2026-03-31
wasm: preserve macro parameters on MACRO lines

ddanila:
## Summary

WASM was expanding symbols too early while reading `MACRO` definitions. If a formal parameter name matched an existing symbol, the recorded macro body captured the global symbol's value instead of the later macro argument.

A blanket no-expand change on `MACRO` lines fixes that bug, but it also regresses defaults such as `p:=CONST` by delaying constant resolution from macro definition time to macro invocation time.

This came up while assembling preprocessed MS-DOS 4.0 CMD sources that rely on MSGSERV macros, where the premature expansion produced incorrect struct-field offsets during later macro expansion.

## Changes

- preserve formal parameter names while recording `MACRO` definitions
- expand only the default-value token span for `:=` parameters when the macro definition is stored
- add focused regressions for both the parameter-capture case and the default-argument case
- keep the new WASM test filenames consistent with the suite's 8.3-style naming

## Impact

Macro bodies now preserve formal parameter references correctly, while default arguments still capture constant values at definition time.

## Validation

- reduced `macroparam.asm` to a smaller reproducer that still fails on upstream `master` and passes on this fix branch
- verified the targeted regressions with `macrparm.asm` and `macrdefl.asm`
- confirmed behavior split with real binaries: upstream `master` misassembles `macrparm.asm` and rejects `macrdefl.asm`, while this branch assembles both correctly
- ran the standalone WASM suite after a clean rebuild: `test1`, `test2`, `test3`, `diag`, and `ideal`


jmalak:
Please, rebase it to last wasm changes, there is new token ":=" to simplify handling of MASM default values and separate it from colon processing.

jmalak:
It looks OK now, Linux WASM tests passed.
After successfully finishing all checks I will submit it.

jmalak:
Please change it from Draft to regular PR to be able submit it.

jmalak:
Thank you for your contribution to WASM.

PR #1597, opened 2026-04-01
wasm: preserve PUBLIC/EXTERN/EXTERNDEF symbol spelling in case-insensitive mode

ddanila:
## Summary

In case-insensitive MASM-compatible mode, `PUBLIC`, `EXTERN`, and `EXTERNDEF` should control the canonical spelling of a symbol when the directive spelling differs only by case. Before this change, WASM kept the first-seen spelling, so OMF `PUBDEF`/`EXTDEF` records could export the wrong case.

This doesn't matter when the linker is also case-insensitive, but becomes significant when linking with a case-sensitive linker or mixing case-sensitive and case-insensitive objects. The fix makes WASM match MASM's behavior: directive spelling wins over first-seen spelling.

This came up while assembling MS-DOS 4.0 CMD utilities, where directives commonly use uppercase while corresponding labels are lowercase.

## Changes

- add `AsmSetMandatoryName()` in `asmsym.c` (standalone-only)
- `PubDef()`, `ExtDef()`: call `AsmSetMandatoryName()`
- add regression tests: `pubcase`, `extcase`, `extcas2`, `extcas3`

## Validation

- standalone WASM test1 suite passes

## Notes

A bit more history could be found [in the fork](https://github.com/ddanila/open-watcom-v2/pull/9).

### DIFF
--- a/bld/wasm/c/asmsym.c
+++ b/bld/wasm/c/asmsym.c
@@ -331,6 +331,17 @@ bool AsmChangeName( const char *old, const char *new )
     return( false );
 }
 
+void AsmSetMandatoryName( asm_sym_handle sym, const char *name )
+/**************************************************************/
+{
+    if( Options.symbols_nocasesensitive ) {
+        if( strcmp( sym->name, name ) != 0 ) {
+            MemFree( sym->name );
+            sym->name = MemStrdupSafe( name );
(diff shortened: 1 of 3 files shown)


jmalak:

PR #1598, opened 2026-04-03
exedump: fix incorrect 'high bound' value for arrays with >256 elements (watcom debug format)

csinkers:
Noticed some incorrect values when investigating Watcom debug data. Pretty simple bug, the array_index function was only ever looking at the first byte of the value, resulting in incorrect results for any arrays with more than 256 elements.


jmalak:

PR #1599, opened 2026-04-05
wdis: fix crash on small data sections in source listing mode

ddanila:
## Summary

`wdis -a` crashes when disassembling OMF objects that contain small data-in-code sections (fewer than 40 bytes). The `tryDUP()` function in `formasm.c` has an unsigned underflow: `size - (8 * MIN_DUP_LINES)` wraps around when `size < 40`, bypassing the guard and causing an out-of-bounds `memcmp` read.

## Reproduction

Any OMF object with a code segment containing a small data region (e.g. a single `DB 0` before instructions) triggers the crash when disassembled with `wdis -a`.

## Fix

Add `size < (8 * MIN_DUP_LINES)` guard before the unsigned subtraction.

### DIFF
--- a/bld/wasmtest/test1/makefile
+++ b/bld/wasmtest/test1/makefile
@@ -80,6 +80,7 @@ shell.dis &
 size.dis &
 sizebug.dis &
 sizlen.dis &
+smalldat.dis &
 stos.dis &
 strmang.dis &
 struct.dis &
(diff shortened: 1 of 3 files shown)


jmalak:
Maybe better to modify the TryDup, to work without 8 * MIN_DUP_LINES hack, but your solution is OK.

PR #1605, opened 2026-04-09
Add netware_clib_lite_dll system for shared NLM modules

playday3008:
`netware_clib_lite` unconditionally links `clibpre.obj` (from the Novell NDK), which references `main` — so you can't build shared library NLMs that have no `main` entry point. All three NDK preludes (`clibpre.obj`, `prelude.obj`, `nwpre.obj`) have this requirement.

Other platforms already have `_dll` variants (`nt_dll`, `os2_dll`, etc.) but NetWare didn't.

This adds a `_THIN_LIB_DLL` guard to `prelude.c` providing `_Prelude` and `_Stop` for shared library NLMs:
- Skips `main()` and `_StartNLM` (no console screen)
- Inits the CRT via `__init_environment` (from `clib3sl.lib`)
- Keeps the NLM resident with a semaphore-blocked keep-alive thread

`predll.c` is a thin wrapper (like `dcinit.c`, `dinitarg.c`, etc.) that includes `prelude.c` with `_THIN_LIB_DLL` defined. Compiled into `predll.obj` for the `ncl` target, referenced by the new `netware_clib_lite_dll` system in `specs.sp`.

Tested by @NishiOwO building [Feather HTTPd](https://github.com/pyrite-dev/feather) shared modules and loading them on NetWare 6.5 SP8.


jmalak:
We always prefer C source code over assembly code, until there is some reason to use assembly code (still can be used C code and for specific code inline assembly)
for this code is no reason why it should be in assembly code because most of external functions and data are in C code and no special code is required and all functions can be simply transform to C, look on bld/clib/startup/c libcstr.c and prelude.c source files
May be it could be shared single C source code for both, shared library and executable and differences can be guarded by some macro
I don't think that C code will be significantly bigger

jmalak:

PR #1607, opened 2026-04-10
Add predll.obj to bld/clib/builder.ctl

NishiOwO:
This should make Builder put predll.obj to lib386/netware, checked working on my machine


jmalak:
please add it to bld/clib/files.dat to be included into OW distribution (installers)

jmalak:
Thanks for your contribution.

PR #1609, opened 2026-04-12
wasm: allow EXTRN:ABS symbols in byte-sized contexts

ddanila:
## Summary

`EXTRN symbol:ABS` in byte-sized contexts incorrectly triggers E050 ("Offset cannot be smaller than WORD size"). This affects both instruction operands (`CMP byte_var, symbol`) and data directives (`DB symbol`). ABS externals are absolute values resolved at link time — the LOBYTE fixup type already exists to handle byte-sized relocations, but E050 guards in both `asmins.c` and `asmdata.c` block ABS symbols from reaching it.

Both MASM 5.10 and TASM 4.1 handle these patterns correctly:

```
$ masm EXTABS.ASM;
      0 Warning Errors
      0 Severe  Errors

$ masm EXTABSDB.ASM;
      0 Warning Errors
      0 Severe  Errors

$ tasm EXTABS.ASM
  Error messages:    None

$ tasm EXTABSDB.ASM
  Error messages:    None
```

## Reproduction

```asm
; Instruction context — E050
.8086
_TEXT SEGMENT BYTE PUBLIC 'CODE'
    ASSUME CS:_TEXT
    EXTRN MY_CONST:ABS
my_var DB 0
    cmp my_var, MY_CONST
_TEXT ENDS
END
```

```asm
; Data directive context — E050
.8086
_TEXT SEGMENT BYTE PUBLIC 'CODE'
    EXTRN MY_CONST:ABS
    DB MY_CONST
_TEXT ENDS
END
```

## Fix

**Instruction path** (`asmins.c`, `idata_fixup`):
- Skip E050 when `opndx->abs` is true (absolute symbol, not a segment-relative offset)
- Add `MT_BYTE`/`MT_SBYTE` cases to the `Code->mem_type` switch so the encoder emits `OP_I8`
- Guard `W_BIT` behind `!MEM_TYPE(Code->mem_type, BYTE)` so byte-context fixups produce correct opcode (80h, not 81h)

**Data directive path** (`asmdata.c`, `array_element`):
- Skip E050 when `init_sym->mem_type == MT_ABS` and emit `FIX_LOBYTE` fixup for the 1-byte case

## Tests

- `bld/wasmtest/test1/extabs.asm` — instruction context (`CMP`)
- `bld/wasmtest/test1/extabsdb.asm` — data directive context (`DB`)

Both fail with E050 on master, assemble and disassemble correctly with this fix.


jmalak:
Thanks for your contribution.

PR #1611, opened 2026-04-14
wasm: fix crash in NAME directive when operand is missing

ddanila:
## Summary

Fix a SIGSEGV in `NameDirective()` when the NAME directive is reached without an operand.

`NameDirective()` reads `tokens[i+1].string_ptr` without checking whether the next token exists. When NAME is reached with no operand (e.g. via the label+directive heuristic rewriting `call name` into label `call` + directive NAME), `tokens[i+1]` is the TC_FINAL sentinel whose `string_ptr` is NULL. This NULL propagates through `MemStrdupSafe` → `ConvertModuleName`, which dereferences it → SIGSEGV.

### Reproducer

```asm
call name
END
```

```
$ wasm -zcm=masm test.asm
test.asm(1): Error! E066: Operand is expected   <-- with fix
Segmentation fault                              <-- without fix
```

### Fix

Add a `TC_FINAL` guard before accessing the token, matching the pattern used by other directive handlers in `direct.c`.

### DIFF
--- a/bld/wasm/c/direct.c
+++ b/bld/wasm/c/direct.c
@@ -4143,6 +4143,10 @@ bool NameDirective( token_buffer *tokbuf, token_idx i )
 /*****************************************************/
 {
     if( Options.module_name == NULL ) {
+        if( tokbuf->tokens[i + 1].class == TC_FINAL ) {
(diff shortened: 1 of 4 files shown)


jmalak:

PR #1614, opened 2026-04-16
wasm: support &macro/&endm keyword escaping in macro bodies

ddanila:
## Summary

Support `&macro` and `&endm` keyword escaping inside macro bodies, enabling nested macro definitions. This is a standard MASM pattern (for example, used in Microsoft's CMACROS.INC from MS-DOS 4.0 sources) that was giving E031 syntax error.

## Changes

* Add `lineis_escaped()` and `lineis_escaped_at()` helpers in `macro.c` that detect `&`-prefixed keywords and strip the `&` prefix
* In the macro body reader, `&endm` is stored as `endm` without affecting nesting depth, and `&macro` is stored as `macro` without incrementing nesting depth
* Add regression test `wasmtest/test1/macroesc.asm`

## Test case

```asm
; Define a macro that creates other macros
defop macro name
op_&name &macro
    nop
&endm
endm

defop foo
defop bar
op_foo
op_bar
```

Verified with MASM 5.10 and TASM 4.1 — identical disassembly output.


jmalak:

PR #1616, opened 2026-04-16
libzip: backport security fixes for CVE-2012-1162, CVE-2012-1163, CVE-2015-2331

ddanila:
Reference: https://github.com/open-watcom/open-watcom-v2/pull/1369

## Description

Backport security fixes to the bundled libzip 0.6.1. A full library upgrade
is not feasible due to breaking API changes in libzip 0.11+ (function renames,
type changes, error handling overhaul in 1.0).

This is the first in a series of PRs splitting the work from #1369 into
separate, reviewable changes -- one library at a time.

## Changes

* **CVE-2012-1162**: Add overflow guard and zero-entry handling in `_zip_cdir_new`
  and `zip_open` to prevent heap buffer overflow when `nentry` is 0 or very large
* **CVE-2012-1163**: Validate `cd->size + cd->comment_len + EOCDLEN` does not
  integer-overflow before use in `fseek` offset calculation in `_zip_readcdir`
* **CVE-2015-2331**: Guard against integer overflow in `sizeof(entry) * nentry`
  allocation in `_zip_cdir_new` (32-bit systems)
* **CVE-2017-12858 (variant)**: Fix memory leaks on error paths in
  `_zip_dirent_read` -- partial allocations (filename, extrafield, comment)
  are now properly freed via `_zip_dirent_finalize` on subsequent failures
* Initialize directory entries before reading in `_zip_readcdir` to prevent
  use of uninitialized memory on cleanup
* Add `extractall.c` regression test simulating the installer workflow

### CVEs assessed but not applicable

* **CVE-2011-0421**: Already patched in the bundled copy (NULL check in `_zip_name_locate`)
* **CVE-2017-14107**: No ZIP64 support in 0.6.1 (`_zip_read_eocd64` does not exist)
* **CVE-2017-12858 / CVE-2019-17582**: The double-free and use-after-free patterns
  from later libzip versions manifest as memory leaks in 0.6.1 due to different
  code structure; addressed by the error-path cleanup above

## Testing

All 5 compilable regression tests from `contrib/libzip/regress/` pass
(compiled with clang on macOS, linked against the patched library and zlib):

* `test_open` -- open/create/error handling for zip archives
* `test_fread` -- reading files (stored, deflated, CRC errors, zlib errors, compressed flag)
* `test_name_locate` -- finding files by name (case-sensitive, case-insensitive, nodir, unchanged flags)
* `test_buffadd` -- add file from buffer, close, reopen with consistency check, read back and verify
* **`test_extractall` (new)** -- simulates installer workflow: open archive, enumerate all entries, stat and extract each one, verify sizes. Also tests round-trip create/add/extract with data integrity check. Uses existing `test.zip` and `broken.zip` test archives.

The remaining 2 test files (`deltest.c`, `ziptest.c`) use a legacy API
(`zip_err_str`, `zip_err`, 2-arg `zip_open`) that predates the bundled
library version and do not compile.

### How to run tests locally

From the repo root:
```sh
# build library and zlib
cc -c -D__UNIX__ -I contrib/libzip/lib -I contrib/zlib -I bld/watcom/h -w contrib/libzip/lib/z*.c
ar rcs /tmp/libzip.a z*.o && rm z*.o
cc -c -D__UNIX__ -I contrib/zlib -w $(ls contrib/zlib/*.c | grep -v -e example -e minigzip)
ar rcs /tmp/libz.a *.o && rm *.o

# build and run a test (e.g. extractall)
cc -D__UNIX__ -I contrib/libzip/lib -I contrib/libzip/regress -I contrib/zlib -I bld/watcom/h \
  -include string.h -include zipint.h \
  contrib/libzip/regress/extractall.c contrib/libzip/regress/mkname.c \
  /tmp/libzip.a /tmp/libz.a -o /tmp/test_extractall
cd contrib/libzip/regress && /tmp/test_extractall
```

### DIFF
--- a/contrib/libzip/lib/zdirent.c
+++ b/contrib/libzip/lib/zdirent.c
@@ -39,7 +39,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
-#include <stddef.h>
 #include <sys/types.h>
 
 #include "wio.h"
@@ -78,7 +77,7 @@ _zip_cdir_new(int nentry, struct zip_error *error)
 {
     struct zip_cdir *cd;
 
-    if (nentry < 0 || (nentry > 0 && (size_t)nentry > SIZE_MAX / sizeof(*(cd->entry)))) {
+    if (nentry < 0 || (nentry > 0 && (size_t)nentry > (size_t)-1 / sizeof(struct zip_dirent))) {
         _zip_error_set(error, ZIP_ER_MEMORY, 0);
         return NULL;
     }
--- a/contrib/libzip/lib/zdirent.c
+++ b/contrib/libzip/lib/zdirent.c
@@ -39,6 +39,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+#include <limits.h>
 #include <sys/types.h>
 
 #include "wio.h"
@@ -77,7 +78,7 @@ _zip_cdir_new(int nentry, struct zip_error *error)
 {
     struct zip_cdir *cd;
 
-    if (nentry < 0 || (nentry > 0 && (size_t)nentry > (size_t)-1 / sizeof(struct zip_dirent))) {
+    if (nentry < 0 || nentry > INT_MAX / (int)sizeof(struct zip_dirent)) {
         _zip_error_set(error, ZIP_ER_MEMORY, 0);
         return NULL;
     }
@@ -90,7 +91,7 @@ _zip_cdir_new(int nentry, struct zip_error *error)
     if (nentry == 0) {
         cd->entry = NULL;
     }
-    else if ((cd->entry=ZIP_ALLOC(sizeof(*(cd->entry))*nentry)) == NULL) {
+    else if ((cd->entry=ZIP_ALLOC(sizeof(*(cd->entry))*(unsigned)nentry)) == NULL) {
         _zip_error_set(error, ZIP_ER_MEMORY, 0);
         ZIP_FREE(cd);
         return NULL;
--- a/contrib/libzip/lib/zdirent.c
+++ b/contrib/libzip/lib/zdirent.c
@@ -39,6 +39,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+#include <stddef.h>
 #include <sys/types.h>
 
 #include "wio.h"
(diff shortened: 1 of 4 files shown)


ddanila:
Converted to draft as I am not confident if backports have not broke something. Library tests are not part of the CI, so... Let me at least run this locally.

UPD: OK, ran some tests, all pass, but still not sure. Feel free to make your verdict. But that's I think is in general problem backporting to the outdated libraries bundled... Left in draft status just to reflect my doubts, otherwise I am done on this PR for now.

jmalak:
It is used by Installer only and tools to create Installer ZIP archive.
The test is possible without breaking by following steps:
- create regular bld/setupgui project, it create csetup/fsetup/setup executable (depend on installer platform).
- download existing installer executable for testing (you need not to create new installer, it is not simple task)
- run csetup/fsetup/setup and add single argument the name of downloaded installer file
- it run new setup executable and use existing installer file as archive
- this way you can debug installer executable too, create debug version of installer executable and run it under debugger with downloaded installer (old version without new changes)

ddanila:
I've also added a.. let's say functional test, cause that was simple; that's just as an addition. Now let me manually verify what you have suggested.

ddanila:
Tested locally against the real OW installer (`open-watcom-2_0-c-linux-x64` from Current-build release, ~123MB):

```
Opening open-watcom-2_0-c-linux-x64...
Archive has 4039 entries
  ...
OK: extracted 4039 entries, 299788167 bytes total
```

All 4039 entries enumerated, stat'd, and extracted (~300MB) with zero errors. This exercises the same libzip code paths as the `setup` binary (`zip_open` → `_zip_readcdir` → `_zip_cdir_new` → `_zip_dirent_read` → `zip_fopen_index` → `zip_fread`).

Also added `contrib/libzip/regress/extractall.c` -- a regression test that simulates the installer workflow using the existing `test.zip` and `broken.zip` test archives.

jmalak:
It is OK, good tests.
You can always do final real test with old installer file.
it needs manual actions, but check everything what is used.
there are some option to do installation sillently that test could be somehow automated.

ddanila:
> It is OK, good tests. You can always do final real test with old installer file. it needs manual actions, but check everything what is used. there are some option to do installation sillently that test could be somehow automated.

But that's what I've done already (see the comment above). Or have I misunderstood you?

contrib/libzip/lib/zdirent.c
jmalak, line 1: You must take into account that code should be compatible with 16/32/64-bit platforms.

The mixing with size_t do new problem, it is not right way.
because nentry is int type then it must be compared as int value (signed).
Because_zip_cdir_new is internal function that such testing is questionable.

it should be something like

`if (nentry < 0 || nentry > INT_MAX / sizeof(struct zip_dirent)) {
`

if you are sure that nentry contains unsigned value only then 

`if ((unsigned)nentry > UINT_MAX / sizeof(struct zip_dirent)) {
`

but then you must cast nentry to unsigned int for size calculation.

This note is valid also for zopen.c changes

ddanila: Yeah, very good catch. I was iterating over these size things, and was not very sure what is correct to use here. Let me try to fix this.

jmalak:
thank you for your contribution.

PR #1617, opened 2026-04-30
wasm: skip macro expansion in IFDEF/IFNDEF/.ERRDEF/ERRIFDEF operands

### LINKED ISSUE #18: WASM: IFDEF/IFNDEF on a name followed later by that name's MACRO definition causes SIGSEGV
## Summary

`wasm` crashes with SIGSEGV (exit 139, no diagnostic output) when an `IFDEF name` or `IFNDEF name` directive appears anywhere earlier in the source than a `name MACRO ... ENDM` definition for the same name. The `IF*DEF` body can be empty — the mere lookup corrupts internal state such that the subsequent macro definition dereferences invalid memory.

## Minimal reproduction

`repro.asm`:

```asm
.8086
.model small
.code
IFDEF foo
ENDIF
foo macro
    nop
endm
end
```

Reproduces on both macOS-arm64 and Linux-x64 in the Apr 20 2026 `Current-build` snapshot.

```
$ wasm repro.asm
$ echo $?
139

$ wasm -zcm=masm repro.asm
$ echo $?
139
```

Crashes in both default and `-zcm=masm` modes. `IFNDEF foo` in place of `IFDEF foo` crashes identically.

## Variations that also crash

- Non-empty `IFDEF` body (e.g., `PURGE foo` inside) — crashes.
- Comments between the `ENDIF` and the `foo macro` line — crashes.
- `IF DEFINED foo` in place of `IFDEF foo` — (not tested; worth checking).

## Variations that do NOT crash

- `IFDEF bar` (different name) followed by `foo macro` — works fine.
- `IFDEF foo / ENDIF` alone with no later `foo macro` — works fine.
- `IF 0 / PURGE foo / ENDIF / foo macro` — works fine (no IFDEF lookup of the macro's own name).

So the trigger is specifically: an `IFDEF`/`IFNDEF` that names a symbol which is later defined as a macro in the same translation unit.

## Impact on MASM compatibility

This blocks the natural MASM idiom for re-includable header files:

```asm
; At the top of a shared INC file:
IFDEF myMacro
    PURGE myMacro
ENDIF
myMacro MACRO ...
    ...
ENDM
```

Such a guard is needed to accompany the PURGE support added in #1615, so that an INC file re-included via different include chains doesn't emit `E236: Procedure is already defined`. With the crash, this pattern is unusable and E236 errors from re-includes cannot be worked around at the source level.

Discovered while migrating MS-DOS 4.0 assembly sources (`dosmac.INC`, `JUMPMAC.INC`, `MSMACRO.INC`, `CHKMACRO.INC`, `RECMACRO.INC`, `FORMACRO.INC`) to WASM; ~137 E236 errors across ~30 targets depend on this pattern being usable.

## Context

- Binary: `Open Watcom x86 Assembler Version 2.0 beta Apr 20 2026 04:18:00 (64-bit)`
- Release tag: `Current-build` (snapshot `ow-snapshot.tar.xz` dated 2026-04-20)
- Platform verified: macOS-arm64 (`armo64/wasm`), Linux-x64 (`binl64/wasm`)

ddanila:
## Summary

Fix [ddanila:#18](https://github.com/ddanila/open-watcom-v2/issues/18) — `IFDEF foo` (or any of `IFNDEF`, `ELSEIFDEF`, `ELSEIFNDEF`, `.ERRDEF`, `.ERRNDEF`, `ERRIFDEF`, `ERRIFNDEF`) followed later by `foo MACRO ... ENDM` crashed `wasm` with a NULL dereference in `strspn`.

## Root cause

On the second assembly pass, `foo` is already in the symbol table as a macro. `ExpandMacro` in `bld/wasm/c/macro.c` scans every line's tokens, spots the macro name, and expands the invocation — even when the identifier is being used as the bare-name operand of a name-argument directive. The expansion replaces the original tokens, so when `get_cond_state` in `bld/wasm/c/condasm.c` later reads `tokbuf->tokens[i+1].string_ptr` for `IFDEF`, it gets the final-token's NULL pointer and `strspn` crashes.

`ExpandMacro` already had this guard for `PURGE` (added by #1615); the bug is that it didn't cover the other directives whose operand is a bare symbol name.

## Fix

Extend the `T_PURGE` skip-expansion check in `ExpandMacro` to include all other bare-name-operand directives: `IFDEF`, `IFNDEF`, `ELSEIFDEF`, `ELSEIFNDEF`, `.ERRDEF`, `.ERRNDEF`, `ERRIFDEF`, `ERRIFNDEF`.

## Reproducer

Minimal repro:

``````asm
.8086
.model small
.code
IFDEF foo
ENDIF
foo macro
    nop
endm
end
``````

Before: `wasm repro.asm` → exit 139 (SIGSEGV), no output.
After: `wasm repro.asm` → 0 warnings, 0 errors.

## MASM 5.10 compatibility — verified

`test1/ifdefmac.asm` was cross-checked against MASM 5.10 (the Microsoft original from `MS-DOS/v4.0/src/TOOLS/MASM.EXE`, Feb 25 1988, version string "@Version=510") running under kvikdos with `--azzy` (relaxed MCB validation for programs that manage their own MCBs):

``````
Microsoft (R) Macro Assembler Version 5.10
Copyright (C) Microsoft Corp 1981, 1988.  All rights reserved.

  48156 + 48094 Bytes symbol space free

      0 Warning Errors
      0 Severe  Errors
``````

MASM 5.10 accepts all the directives exercised by the test (`IFDEF`, `IFNDEF`, `ELSEIFDEF`, `ELSEIFNDEF`, `.ERRNDEF`) and produces a valid `IFDEF.OBJ`. The canonical MASM re-include guard —

``````asm
IFDEF name
    PURGE name
ENDIF
name MACRO ...
    ...
ENDM
``````

— uses `IFDEF`, `PURGE`, `MACRO`/`ENDM`, all documented MASM 5.x features and used throughout MS-DOS 4.0's own `dosmac.INC`, `JUMPMAC.INC`, `MSMACRO.INC`, and `cmacros.inc`. The fix aligns WASM's behaviour with MASM 5.10 on that pattern.

## Test plan

- [x] `test1/ifdefmac.asm` — covers `IFDEF`, `IFNDEF`, `ELSEIFDEF`, `ELSEIFNDEF`, `.ERRNDEF`, each with `foo` as operand while `foo` is also defined as a macro.
- [x] Existing `test1/purge.asm` (the `T_PURGE` case, unchanged behavior) still assembles cleanly — covered by construction since the switch preserves `T_PURGE`.
- [x] Existing `test1/macroesc.asm` still assembles cleanly (unrelated but verified against the rebuilt wasm).
- [x] `ifdefmac.dis` added to `bld/wasmtest/test1/makefile` so the regression is exercised on every run.
- [x] Same test file assembles cleanly under MASM 5.10 (0 warnings, 0 errors).

`T_DOT_ERRDEF` / `T_ERRIFDEF` cannot be tested cleanly (they fire whenever their argument is defined, which is exactly the repro condition). `T_ERRIFDEF` / `T_ERRIFNDEF` are TASM-only in default WASM mode. All three share the same switch case as the tested directives.


jmalak:
Something is wrong, below is test results

==== 19:18:13 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test1 ====
TEST 386/00 (64-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test1
==== 19:18:14 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test1 ====
TEST 386/00 (32-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test1
==== 19:18:15 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test2 ====
TEST 386/00 (64-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test2
==== 19:18:16 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test2 ====
TEST 386/00 (32-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test2
==== 19:18:17 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test3 ====
TEST 386/00 (64-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test3
==== 19:18:18 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test3 ====
TEST 386/00 (32-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/test3
==== 19:18:18 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/ideal ====
TEST 386/00 (64-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/ideal
==== 19:18:18 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/ideal ====
TEST 386/00 (32-bit)
PASS /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/ideal
==== 19:18:18 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/diag ====
TEST 386/00 (64-bit)
==== 19:18:19 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest/diag ====
TEST 386/00 (32-bit)
==== 19:18:19 /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest ====
TEST
FAIL /home/runner/work/open-watcom-v2/open-watcom-v2/bld/wasmtest

It looks like diagnostics tests failed for some reason.

ddanila:
Thanks for catching this. Tracked it down — the failing test is `wasmtest/diag/namedir.asm`, and the bug is actually in commit d8434a98 / #1611, not this PR.

`NameDirective()` guards the missing-operand check inside `if( Options.module_name == NULL )`. The diag test framework invokes wasm with `-nm=$[&` for every test, so `Options.module_name` is already set when `call name` is parsed (label `call` + `NAME` directive with no operand) — the `TC_FINAL` guard is skipped and `NAME` silently returns `RC_OK` instead of emitting E066. The diff against `namedir.chk` then fails. The test has been failing since #1611 was merged.

I've folded a fix into this PR — moves the `TC_FINAL` guard above the `module_name == NULL` check so E066 is emitted regardless of how `module_name` was set.

Side note for visibility: this isn't surfacing in CI because `wmake -h` returns non-zero on diag's failure but the wasmtest action doesn't propagate the exit status — the job log ends with `<wmake -h test0=1 host=386> => non-zero return: 512` but the step still reports green. Might be worth a separate issue.

jmalak:
Be careful, regression tests are work-in-progress that don't cause OW build fail.
It is necessary check tests log to determine what fails.
Because some tests can be wrong after fixing some problem in WASM that if it happen it is necessary to check OW code base if not use similar wrong construct etc.

Now All is OK.
Thanks for your contribution.

PR #1618, opened 2026-05-02
Fix wasm struct label initialization

ddanila:
## Summary

Fixes a wasm crash when expanding a structure initializer for a structure that contains a `LABEL` member.

`LABEL` members inside structures create an offset alias and do not emit data. They can therefore have no initializer text in the structure field list. `InitializeStructure()` previously assumed every field had initializer text and called `strlen()` unconditionally, which dereferenced null for these label-only fields.

This change skips label-only fields during structure initializer expansion, preserving their offset symbol while avoiding both data emission and initializer consumption.

## Regression Test

Adds `bld/wasmtest/diag/structlb.asm`, a minimal structure with a `label byte` member and an initializer that previously crashed wasm.

## Validation

Ran locally:

```sh
. ./setvars.sh >/tmp/ow-setvars.log && cd bld/wasmtest/diag && rm -f structlb.obj structlb.out PASS && wmake -h buildtest=1 arch=386 structlb.dum
```

The focused diagnostic target passes.

### DIFF
--- a/bld/wasm/c/asmstruc.c
+++ b/bld/wasm/c/asmstruc.c
@@ -166,6 +166,13 @@ bool InitializeStructure( asm_sym_handle sym, asm_sym_handle struct_symbol, toke
         /* put the lines to define the fields of the structure in,
          * using the values specified ( if any ) or the default ones otherwise
          */
+        if( f->initializer == NULL ) {
+            /*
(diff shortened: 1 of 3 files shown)


jmalak:
Thanks for your contribution.

PR #1619, opened 2026-05-03
buildx.sh: Update comments in the CI script

winspool:
Building OpenWatcom on the Azure Pipelines has ended long ago.
Also fixes the Typo nearby. 

There is no other location of "Azure" in the OpenWatcom sources.


jmalak:
Thanks for your contribution.

PR #1620, opened 2026-05-03
ras.mh: Fix typo in RAS_PROJECTION_INFO

winspool:
Found this typo when fixing the typo in ci/buildx.sh

Verified this change with:
https://learn.microsoft.com/en-us/windows/win32/api/ras/ns-ras-ras_projection_info


jmalak:

PR #1621, opened 2026-05-07
wasm: don't join trailing-comma lines outside WATCOM mode

ddanila:
## Summary

In MASM-compatible mode (`-zcm=masm`) `wasm` eats the directive that follows a macro invocation ending with a trailing comma. The trailing comma is supposed to leave the macro's last positional parameter blank — that's what MASM and TASM both do — but `wasm` instead joins the next physical line as continuation, silently consuming the `END` (or similar) directive and bailing with E249 *End directive required at end of file*.

## Repro

```asm
        .286
        .model  small
        .code

mymacro macro a, b
        nop
endm

        mymacro 1,
        end
```

```
$ wasm -zcm=masm -0 -ms repro.asm
repro.asm(10): Error! E249: End directive required at end of file
```

After fix: 0 errors, OBJ produced.

## Cause

`bld/wasm/c/asmline.c` — `join_multiline_cmds()` unconditionally appends the next physical line whenever the current line's last non-whitespace character is `,`. That's a WATCOM-mode extension. MASM and TASM treat the trailing comma as a complete statement with the trailing positional argument left empty.

## Fix

Gate the comma-continuation on `(Options.mode & MODE_WATCOM)`:

```c
if( (Options.mode & MODE_WATCOM) == 0 )
    return( line );
```

WATCOM-mode behavior is unchanged.

## Cross-checked against the assemblers wasm emulates

| Assembler | Result |
|---|---|
| Microsoft Macro Assembler **5.10** | accepts cleanly, OBJ produced |
| Turbo Assembler **4.1**            | accepts cleanly, OBJ produced |
| `wasm -zcm=masm` (before fix) | E249 |
| `wasm -zcm=masm` (after fix) | accepts, OBJ produced |
| `wasm` (default WATCOM mode, after fix) | unchanged |


jmalak:
It looks like comma on end-of-line should be continuation for wasm and MASM 6.1 mode.
That check should be for MODE_MASM6 and MODE_WATCOM or MODE_MASM5 and MODE_TASM
I am not sure if it is general rule or only in some context.

jmalak:

PR #1622, opened 2026-05-07
wasm: macro substitution drops quote delimiters from TC_STRING args

ddanila:
## Summary

When a single- or double-quoted character argument is substituted into a macro body, `wasm` drops the surrounding `'`/`"` delimiters from the substituted text. So `m '/'` with body `DB x` substitutes `x -> /`, the body becomes `DB /`, and `wasm` rejects with `E020 "Expecting number"`.

## Repro

```asm
        .286
        .model  small
        .data

m macro x
        DB x
endm

        m '/'
        end
```

```
$ wasm -0 -ms repro.asm
repro.asm(7): Error! E020: Expecting number
```

## Fix

Two substitution modes, picked by reference style at preprocess time:

* **Standalone reference** (e.g. body `DB x`, `m '/'` -> `DB '/'`): re-wrap the substituted content with the arg's original `'`/`"` delimiter, doubling embedded delim chars. `<...>` keeps dropping delimiters (raw text, what STRUC.INC depends on).
* **`&`-adjacent reference inside a string** (e.g. body `'&cl'` in MS-DOS `CMACROS.INC`'s `createSeg`, called as `createSeg ...,CODE,...` -> `'CODE'`): `&` is a text-substitution operator that splices the arg's text into the surrounding string. The arg's outer quotes (if any) drop -- this is `wasm`'s historical behavior, on which `bld/watcom/h/mdef.inc`'s `extrn "&p1",`p2`` (called as `xrefn "C",__set_EINVAL`) depends.

Implementation:
* `asm_tok` gets a `string_delim` field set by `get_string` from the opener.
* `parm_list` carries the arg's delim per invocation, recorded only for a single-`TC_STRING`-token arg.
* `replace_param` emits `#dd` (wrap on substitute) or `@dd` (raw on substitute); `my_sprintf` dispatches on the kind.

Regression tests:
* `bld/wasmtest/test1/quotmac.{asm,esm}` -- standalone-reference round-trip across `'/'`, `"ab"`, `"x'y"`, `'a''b'`, `<99>`.
* `bld/wasmtest/test1/qmacand.{asm,esm}` -- `&p` inside body string across bare / `"..."` / `'...'` / `<...>` args.

## Cross-checked against the assemblers wasm emulates

Run under MS-DOS 4.0 in QEMU, against MASM (ML 6.x) and TASM 4.1.

| Case | MASM | TASM | `wasm` before | `wasm` after |
|---|---|---|---|---|
| `m '/'` -> `DB x` round-trip | accepts | accepts | E020 | accepts |
| `m AB` -> `DB '&p'` (bare arg, body wraps -- MS-DOS `createSeg`) | accepts (`'AB'`) | accepts (`'AB'`) | accepts | accepts |
| `m "Z"` -> `DB "&p"` (quoted arg, body re-wraps -- `mdef.inc`) | A2206 | "Extra characters on line" | accepts | accepts |

The third case is `wasm`-specific: MASM/TASM substitute the arg's text *with quotes*, producing `DB ""Z""` (which they reject); `wasm` strips the outer quotes, producing `DB "Z"`. Preserving this divergence keeps `mdef.inc` working.


ddanila:
Unfortunately, was not able to make this small enough ( Initial naive approach failed. But the scenario that I am trying to support here is real and from the msdos 4.0 sources. Will also investigate the windows build failure; no idea how it could be related so specifically with my changes yet.

jmalak:
I will check, but may be to we will need to add / (slash) character as parser token.

ddanila:
OK, windows build passed now...

jmalak:
Probably right solution will be to define new token classes TC_STRING1-TC_STRING... instead of TC_STRING in enumaration constants `enum tok_class` and existing wasm code will need more review for this processing with regards to real string is using only single/double quotes and other forms is mostly context specific.

ddanila:
Yeah, overloading existing TC_STRING with too much metadata is definitely a wrong way, totally agree. The only argument doing in that way was to keep PR small; but it grew up supporting all edge cases anyway.

jmalak:
It is OK. I also don't know each functionality in detail that I need to check and think about ideas.
I think that original design was too simple (MASM 5.x only) to get such difference and fixing other related problems need more changes, it is also reason for minimal diagnostics.
Lexer should hold TC_STRING for single/double quoted strings only and for all other cases not use TC_STRING at all but `TC_<special_open token> <item token type> TC_<special_close token>` and modify all directives to handling this change. I understand it is much big change, but there will be more trouble if we don't resolve root cause. We can do quick fix and next do this design change which need more effort and time.

ddanila:
Yep, let me try to craft something.

jmalak:
I checked how complex is lexer change and it looks like we should use your change.which use multiple TC_STRING_... as you prepare.
Lexer change is more significant and will need big rework of wasm code especialy expression evaluation.

ddanila:
> I checked how complex is lexer change and it looks like we should use your change.which use multiple TC_STRING_... as you prepare. Lexer change is more significant and will need big rework of wasm code especialy expression evaluation.

Totally up to you. Your approach looks doable as well, the PR would be large, but not huge... Like 2-3 times larger than currently.

At the same time, last TC_STRING* approach would not hurt much, we can always start refactoring on top of it merged.

jmalak:
I will submit it as first solution. In case we have more trouble then we can refactore it.

PR #1644, opened 2026-08-02
cg: fix RISC subregister mapping

hakanrw:
The RISC register-table lookup used HW_Equal() to match register sets
against canonical entries in RegsTab.

Narrow integer registers use byte and word register sets that overlap
their canonical physical register but are not exactly equal to it.
MIPS 64-bit values may similarly use paired register sets. Lookup
therefore failed for these register forms and GetArchIndex() returned
register 0.

On MIPS, narrow loads could consequently target the read-only $zero
register and silently discard their result.

Reproducer:
```
    unsigned short narrow_word(unsigned long value)
    {
        return value;
    }
```

Compile with:
```
    wccmps -od test.c
```

Before:
```
    narrow_word:
        sw      $a0,($sp)
        lhu     $zero,($sp)
        andi    $v0,$zero,0xffff
        jr      $ra
        nop
```

After:
```
    narrow_word:
        sw      $a0,($sp)
        lhu     $v0,($sp)
        andi    $v0,$v0,0xffff
        jr      $ra
        nop
```

The fix uses overlap-aware lookup for the MIPS, Alpha, and PowerPC
register tables. This maps byte, word, full-width, and paired register
sets to their physical architectural register. An unrecognized register
now reports an internal compiler error instead of silently selecting
register 0.

### DIFF
--- a/bld/cg/risc/axp/c/axprgtbl.c
+++ b/bld/cg/risc/axp/c/axprgtbl.c
@@ -856,48 +856,41 @@ void            InitRegTbl( void )
 {
 }
 
+static int findArchRegInfo( hw_reg_set regs )
+/********************************************/
+{
+    int         i;
+
+    for( i = ARCH_IDX_START; i < ARCH_IDX_END; i++ ) {
+        if( HW_Ovlap( regs, RegsTab[i].hw_reg ) ) {
+            return( i );
+        }
+    }
+    _Zoiks( ZOIKS_031 );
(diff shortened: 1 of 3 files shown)


jmalak:

jmalak:
Be carefull if you disassembly generated code, OW disassembler can be wrong, it is work-in-progress because tools for RISC are untested.
I think there are some option to use register number instead register alias to be sure if register number is OK.

PR #1647, opened 2026-08-10
wl: fix MIPS PE jump base relocations

hakanrw:
Wlink emitted `HIGHLOW` PE base relocations for MIPS 26-bit jump
relocations.

When an image was loaded away from its preferred base, the loader added
the complete byte delta to the instruction word. This corrupted `j` and
`jal` targets and could cause an immediate crash. Whether a program
worked therefore depended on whether its preferred image base was
available.

Reproducer:

```
    int main(void)
    {
        return 0;
    }
```

Compile and inspect the relocation table:

```
    wclmps test.c
    wdump -f test.exe
```

The entry point is at RVA `0x10020`, with a `jal` relocation at
`0x10028`.

Before:

```
    3:028     3:058     4:060      0650     3:080     3:098
```

After:

```
    5:028     5:058     4:060      0650     5:080     5:098
```

Type 3 is `HIGHLOW`; type 5 is `MIPSJMPADDR`.

The fix maps MIPS `FIX_OFFSET_26` relocations to the existing
`PE_FIX_MIPSJMP` relocation type. Other architectures and relocation
kinds are unchanged.


jmalak:
Thanks for your contribution

Issue #1582, opened 2026-02-13
OS/2 builds instantly crash with SYS3175


mrpapersonic:
The `Current-build` release instantly crashes:

<img width="543" height="466" alt="Image" src="https://github.com/user-attachments/assets/86b4f1e4-df35-4a7d-a5cd-4dc5db7553c5" />

mrpapersonic:
I think *all* builds made with recent versions of OWv2 also have this problem. I believe it started happening somewhere between May and July of last year.

vain:
Hi @jmalak,

I also noticed that my OS/2 builds were broken. And like @mrpapersonic said, this means *all* builds, not just the installer. Cross-compiling a simple hello world from Linux yields the same breakage:

```c
#include <stdio.h>

int
main()
{
    printf("Hello, OS/2!\n");
    return 0;
}
```
```
owcc -Wall -Wextra -std=c99 -bos2 -o test.exe test.c
```

I bisected this and found that the last good commit is 2e1e754e60 and the first bad commit is b4ecde66d9. There are a few commits in between (but not many), but none of them build for me or the compiler crashes.

Here are the compiled EXE files: https://github.com/vain/open-watcom-os2-sys3175

I hope this helps finding the bug. :)

Thank you!

jmalak:
Thanks for your investigation of issue.
I will look on this issue, which is for OS/2 executable.
The problem is that we have not regular test for OS/2 build. We use now Linux, Windows or OSX build only which are supported by Github Actions (Workflow).
I do testing time-to-time on my VM only not on regular base.
Please, what target you want to build, because your owcc command is for build 16-bit OS/2 executable not for 32-bit executable.
To build 32-bit OS/2 executable you need to use **`-bos2v2`** option instead of **`-bos2`**.

mrpapersonic:
I do builds for 32-bit OS/2 w/ PM. The broken builds' timestamps do seem to line up with the commits vain mentioned. 16-bit OS/2 may be broken as well, but I don't build for it, so I can't add any input there.

jmalak:
if you build for 32-bit then you must use **`-bos2v2`** option for owcc
anyway it should works for both 16 and 32-bit target.

jmalak:
I checked the changes  in you defined range and the issue could be related to changes which introduced __noreturn modifier to optimize CRTL.

vain:
[...]

PR #1649, opened 2026-08-11
clib: fix OS/2 32-bit startup arguments

hakanrw:
The 32-bit OS/2 CRT entry point was incorrectly marked `_WCNORETURN`.

For OS/2 executable entry points, the loader-provided stack contains a
return address followed by the module handle, reserved value, environment
pointer, and command-line pointer. The `_WCNORETURN` annotation changed
the compiler parameter offsets so that the return-address slot was not
accounted for, which breaks OS/2 ABI.

Thus, `__OS2Main()` received shifted arguments and environment pointer 
was interpreted as the command-line pointer. `__OS2MainInit()` then 
searched backwards from that address for the program name and 
would read outside accessible memory.

Reproducer:

```
    #include <stdio.h>
    
    int main(void)
    {
        printf("Hello from OS/2\n");
        return 0;
    }
```

Compile with:

```text
    wcl386 -bt=os2 -l=os2v2 test.c
```

Before the fix, launching the executable produced a SYS3175 access
violation during CRT startup, before `main()` was reached.

After the fix, the console application starts, prints its output, and
exits normally.

The fix removes `_WCNORETURN` from the 32-bit OS/2 entry-point
declarations and definition, restoring the loader-compatible parameter
layout. The 16-bit OS/2 startup path is unchanged.

The change was tested under OS/2 Warp 3.0.

### DIFF
--- a/bld/clib/startup/c/main2o32.c
+++ b/bld/clib/startup/c/main2o32.c
@@ -52,7 +52,7 @@
 #include "xmain.h"
 
 
-_WCNORETURN void _WCNEAR __F_NAME(__OS2Main,__wOS2Main)( unsigned hmod,
+void _WCNEAR __F_NAME(__OS2Main,__wOS2Main)( unsigned hmod,
                             unsigned reserved, char *env, char *cmd )
 /********************************************************************/
 {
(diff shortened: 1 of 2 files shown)


jmalak:
Thanks for fix, I have had no-time to fxi it, issue #1582

PR #1650, opened 2026-08-11
clib: enable OS/2 32-bit default windowing

hakanrw:
Open Watcom contains an OS/2 Presentation Manager implementation of
default windowing, but the feature was not enabled by the OS/2 32-bit
clib build.

This enables default windowing for the i386 OS/2 target and fixes the
remaining compilation and teardown issues in that path.

During shutdown, `_MainWindowData` was freed before the PM windows were
destroyed. Window destruction dispatches messages through
`_MainDriver()`, which still requires that data. The cleanup order is
changed to destroy the window first and release its bookkeeping
afterwards.

Reproducer:

```
    #include <stdio.h>
    
    int main(void)
    {
        printf("Hello from OS/2!\n");
    
        printf("\nPress Enter to exit...");
        getchar();

        return 0;
    }
```

Build as a normal window-compatible VIO application:

```
    wcl386 -bt=os2 -l=os2v2 main.c -fe=main-vio.exe
```

Build using Open Watcom default windowing:

```
    wcl386 -bt=os2 -bw -l=os2v2_pm main.c -fe=main-bw.exe
```

The normal build runs in an OS/2 VIO window. The `-bw` build creates the
Open Watcom PM frame and its `Standard IO` child window, with
`printf()` and `getchar()` redirected through the existing default
windowing infrastructure.

VIO:
image

OW DW:
image

Both modes were tested on OS/2 Warp 3.0, including input and clean
program termination.

This change enables default windowing only for 32-bit i386 OS/2. The
16-bit OS/2 path remains unchanged.

### DIFF
--- a/bld/clib/defwin/c/pmmain.c
+++ b/bld/clib/defwin/c/pmmain.c
@@ -37,6 +37,8 @@
 #define INCL_WIN
 #include <wos2.h>
 #include "win.h"
+#include "defwin.h"
+#include "initwin.h"
 #include "pmmenu.rh"
 
 
@@ -189,10 +191,10 @@ _WCRTLINK void  __InitDefaultWin( void )
 _WCRTLINK void  __FiniDefaultWin( void )
 //======================================
 {
-    _FiniMainWindowData();
     if( _MainWindow != 0 ) {
         WinDestroyWindow( _MainWindow );
     }
+    _FiniMainWindowData();
     if( hMessageQueue != 0 ) {
         WinDestroyMsgQueue( hMessageQueue );
     }
--- a/bld/clib/defwin/h/win.h
+++ b/bld/clib/defwin/h/win.h
@@ -53,7 +53,7 @@
 #ifdef __OS2__
 typedef int         HANDLE;
 extern  unsigned    __WinSetWindowPos(unsigned);
-#pragma aux __WinSetWindowPos __parm [__eax] __modify [__ebx]
+#pragma aux __WinSetWindowPos = __parm [__eax] __modify [__ebx]
 #define WinSetWindowPos(a1,a2,a3,a4,a5,a6,a7)           \
         __WinSetWindowPos(WinSetWindowPos(a1,a2,a3,a4,a5,a6,a7))
 #endif
--- a/build/mif/defwind.mif
+++ b/build/mif/defwind.mif
@@ -1,3 +1,8 @@
 !ifeq system windows
 default_windowing = 1
 !endif
+!ifeq system os2
+!ifeq host_cpu 386
+default_windowing = 1
+!endif
+!endif


hakanrw:
CI failed, I am checking the build error currently. I will update 
the PR branch after fixing the issue.

jmalak:
The default Windowing is used only for 16-bit Windows (+32-bit Win Extender), because it doesn't support console ( command line).
For another OS-es it is deactivated because they have console (command line), but should be compilable. It can be tested if 
you set environment variable OWDEFAULT_WINDOWING=1 on OS/2 and NT.

hakanrw:
CI exposed that the OS/2 387 runtime variant defines a different processor,
which gated the default windowing for that variant and failed the build 
while trying to reuse `nodefwin.obj` (which didn't exist because 386 had
built `dinitwin.obj` instead).

I updated the default-windowing gate to use `host_cpu == 386` and 
verified both 386 and 387 runtime DLL builds.

> The default Windowing is used only for 16-bit Windows (+32-bit Win Extender), because it doesn't support console ( command line). For another OS-es it is deactivated because they have console (command line), but should be compilable. It can be tested if you set environment variable OWDEFAULT_WINDOWING=1 on OS/2 and NT.

Oh, I see, I did not know that environment flag. Enabling it would have failed 
previously, however, because the OS/2 default-windowing sources contained 
compilation errors. This PR also fixes those issues.

If default windowing is intentionally disabled for OS/2, I can drop the .mif 
change that enables it by default while preserving the compilation and teardown 
fixes, so OWDEFAULT_WINDOWING=1 remains usable for testing.

jmalak:
> CI exposed that the OS/2 387 runtime variant defines a different processor, which gated the default windowing for that variant and failed the build while trying to reuse `nodefwin.obj` (which didn't exist because 386 had built `dinitwin.obj` instead).
> 
> I updated the default-windowing gate to use `host_cpu == 386` and verified both 386 and 387 runtime DLL builds.
> 
I'm not sure if I understand you. CRTL doesn't use any definition of FPU. 
Could you elaborate a bit more please.

jmalak:
Only note to your change
`#pragma aux __WinSetWindowPos = __parm [__eax] __modify [__ebx]`
the equal character has no sense here
equal character has sense to be used if you add some instructions code after it, it must be instruction  bytes order.
The #pragma aux is very flexible and can be used for code generator low-level symbol information definition or to overwrite default info
The original form is OK, because without equal char it is transparent that it is not code but symbol properties only
normally **`"default"`** **`#pragma aux`** properties are used until it is changed by explicit **`#pragma aux`** specification

hakanrw:
> > CI exposed that the OS/2 387 runtime variant defines a different processor, which gated the default windowing for that variant and failed the build while trying to reuse `nodefwin.obj` (which didn't exist because 386 had built `dinitwin.obj` instead).
> > I updated the default-windowing gate to use `host_cpu == 386` and verified both 386 and 387 runtime DLL builds.
> 
> I'm not sure if I understand you. CRTL doesn't use any definition of FPU. Could you elaborate a bit more please.

Sure, I was not clear on this one. Let me explain in detail,

The failing target was the OS/2 wr7 RTDLL build. It sets `processor=387` to
select the x87-specific math387, but it reuses the regular OS/2 386 CRTL and 
its objects.

The CRTL was built with `processor=386`, so the original condition enabled 
default windowing and produced `dinitwin.obj`. Non-x87 wrt RTDLL linked
correctly. The later wr7 build used `processor=387`, did not enable default 
windowing, and the build config selected `nodefwin.obj` from the same CRTL 
object directory.

Since the earlier CRTL build had produced `dinitwin.obj` instead, 
`nodefwin.obj` did not exist and the build failed. This is my understanding of
the CI fail.

Both variants map to `host_cpu=386`, so I changed the condition to use
`host_cpu` instead of `processor` and verified both RTDLL builds, the CI passes
now.

However, I now plan to remove the automatic enabling of default windowing 
on OS/2, reverting the `.mif` changes while retaining the compilation and 
teardown fixes. That will make this makefile distinction irrelevant to the final 
change

> Only note to your change `#pragma aux __WinSetWindowPos = __parm [__eax] __modify [__ebx]` the equal character has no sense here equal character has sense to be used if you add some instructions code after it, it must be instruction bytes order. The #pragma aux is very flexible and can be used for code generator low-level symbol information definition or to overwrite default info The original form is OK, because without equal char it is transparent that it is not code but symbol properties only normally **`"default"`** **`#pragma aux`** properties are used until it is changed by explicit **`#pragma aux`** specification

Thank you for the feedback. I'll take another look at this.

jmalak:
It is clear for me now what you mean by 387.
I think it is not necessary to change it now, due to we not support default windowing for 32-bit OS/2 and NT.
Anyway wrt DLL is work-in-progress, not yet tested, only build to be sure not broken, but not part of distribution.
The idea of WATCOM was probably to consolidate all partial run-time libraries to single big run-time DLL, I am not sure if it is better then 3 run-time libraries (clib, cpplib and mathlib). I didn't try to build OW with this run-time DLL because it requires to change build make files and rebuild OW and test.
Anyway if you have any question then ask me and if I know I will answer. There is many specific things due to multiplatform nature of OW that it can look oddly. 
Take into account that OW can be build on following hosts DOS, OS/2, NT, Linux, OSX and BSD class, of cause not fully, it is possible on NT only. Now it is tested only on Linux, OSX and NT.

jmalak:
additional info for wrt DLL build in bld/rtdll/master.mif is remapping 387 to 386 that you need not to do it separately
only there can be something missing or incorrect macro if it doesn't work

jmalak:
I little reworked it and fix other problems related to default Windowing stuff
submitted as commit 340c139c15c3fa576c3a742128a10be21bb876d5

hakanrw:
Hi Jiri, thanks! I had not had the time to look into this, thanks for sorting it out.

jmalak:
no problem, i test it and fix all bugs for NT and OS/2 default windowing, which is not important (unused now)

PR #1656, opened 2026-08-17
Add IBM HLL debug type support

komh:
Hi/2.

This PR is to add IBM HLL debug type support to Open Watcom V2.

Review, please...


jmalak:
Thanks for your contribution.

Please use following rules to format source code
1. use always spaces, not hard tabs
2. remove all date stamps and authors (it is hold by SCM)
3. use types specific constants
- pointers NULL
- string termination character '\0'

Rebase to the latest git source code

komh:
I think, I've fixed according to the rules.

jmalak:
thanks for your update.

PR #1657, opened 2026-08-18
cg: add OWL COFF and ELF object output for 386

hakanrw:
This adds an MVP OWL object writer alongside the existing Intel OMF writer.

The original `x86obj` implementation is renamed to `x86omf`. A new `x86obj` layer dispatches the existing object-output operations through an emitter table. This interface is OMF-shaped: existing Intel CG callers remain the same, and the OMF implementation remains the default.

For 386 targets:

- `-eoo` selects OMF.
- `-eoc` selects OWL/COFF.
- `-eoe` selects OWL/ELF.

The switches are available in both the C and C++ frontends. The RISC C++ frontends also gain their existing OWL COFF/ELF selections.

The OWL path currently supports:

- Flat 32-bit code, data and BSS sections.
- Local and global symbols.
- Function and object symbols.
- Absolute and PC-relative 32-bit relocations.
- Calls and references to local and external symbols.
- Compiler-local label patches.
- Imports, exports, aliases, default libraries and dependency information.
- COFF and ELF output.

The following are not yet implemented:

- Debug information.
- 16-bit sections or segmented/far addressing.
- TLS and `far16` relocations.
- Non-offset relocation classes.
- Floating-point patch records.
- Conditional/common import fixups.
- Virtual-function linker directives.

Unsupported paths report an explicit fatal error and bail out without emitting an object.

### Review guidance

I split the commits in intended review order:

1. Rename of the original `x86obj` writer to `x86omf`.
2. Addition of the emitter interface and dispatcher. The OMF table should map one-to-one to the previous implementation.
3. Addition of the OWL implementation. Each interface operation can be compared directly with its OMF counterpart.
4. ELF x86 PC-relative addend handling.
5. C++ frontend object-format switches.
6. ~~Fix on build dependencies for generated command-line parser and help sources.~~ (Dropped)

The existing OMF path can be checked with `-eoo` or with no object-format option. Basic source files can then be compiled with `-eoc` and `-eoe`, and the resulting symbol tables, sections and relocations compared across the three formats.


bld/plusplus/master.mif
jmalak, line 1: this explicit dependency is useless, dependency of these files are resolved indirectly by macro prebuild_objs

bld/cc/master.mif
jmalak, line 1: this explicit dependency is useless, dependency of these files are resolved indirectly by macro prebuild_objs

hakanrw:
My bad, sorry! Apparently it was a build issue on my end and I overcorrected it.

I removed the last commit that introduced the change in mif config.

jmalak:
No problem, I understand you can not know such details about OW build system.
It is based on build/mif/...mif make files which hold rules and use some macros to simplify complex multiplatform rules.
it is done this way for all project make files if there is some .gc, .gh, .grh any files with extension .g... then it is generated files and mostly it is handled by prebuild_objs macro.

